PR #3, opened 2017-05-27
Added static casts of `size_t` container sizes

w1th0utnam3:
Previously `size_t` values returned by the internal containers were implicitly converted to the ECS `size_type` which results in compiler warnings about possible loss of accuracy (if `size_type` < `size_t`). If you want to make the user aware of this issue, I would rather add an assert in the `create()` method or something like this.

### DIFF
--- a/src/component_pool.hpp
+++ b/src/component_pool.hpp
@@ -50,11 +50,11 @@ public:
     }
 
     size_type capacity() const noexcept {
-        return data.capacity();
+        return static_cast<size_type>(data.capacity());
     }
 
     size_type size() const noexcept {
-        return data.size();
+        return static_cast<size_type>(data.size());
     }
 
     const entity_type * entities() const noexcept {
@@ -82,7 +82,7 @@ public:
             reverse.resize(entity+1);
         }
 
-        reverse[entity] = direct.size();
+        reverse[entity] = static_cast<size_type>(direct.size());
         direct.emplace_back(entity);
         data.push_back({ args... });
 
--- a/src/registry.hpp
+++ b/src/registry.hpp
@@ -246,11 +246,11 @@ public:
     Registry & operator=(Registry &&) = delete;
 
     size_type size() const noexcept {
-        return count - available.size();
+        return static_cast<size_type>(count - static_cast<entity_type>(available.size()));
     }
 
     size_type capacity() const noexcept {
-        return count;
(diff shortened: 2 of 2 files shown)


skypjack:
At the end of the day, the type of the entities rules on the size and the size type. I added `Entity` template parameter and forgot to clean up that part. I don't get this PR solves the issue, it's simply a workaround. Let me review it entirely and propose something different.

w1th0utnam3:
I was a little bit confused why `size_type` is currently fixed to `std::uint32_t`. And yes, obviously just a workaround - probably not sensible for this general repository in comparison to the toy project I use it for.

skypjack:
I pushed on master a few changes that should clean things up. Can you give me a feedback, please?

w1th0utnam3:
1. I get the internal compiler error of MSVC again in the base class initialization of `ComponentPool` ([here](https://github.com/skypjack/entt/blob/73501809b4b621aca7c3446ba77593c812a2dc27/src/component_pool.hpp#L127)). I have to experiment to find a workaround.
2.  There is still a potentially lossy conversion [here](https://github.com/skypjack/entt/blob/73501809b4b621aca7c3446ba77593c812a2dc27/src/component_pool.hpp#L81) (`std::size_t` to `pos_type`)

skypjack:
Oh, I'm sorry, appveyor is having problems for I didn't set master as default branch. I'm going to fix it, not a problem. Give me a couple of minutes.

w1th0utnam3:
The problem is the usage of the alias template
```
template<typename Comp>
using Pool = ComponentPool<Entity, Comp>;
```
in the pack expansion context of the initializer:
```
explicit ComponentPool(size_type dim = 4098) noexcept
        : Pool<Component>{dim}, Pool<Components>{dim}...
```
Replacing it with the full class name:
```
explicit ComponentPool(size_type dim = 4098) noexcept
        : ComponentPool<Entity, Component>{ dim }, ComponentPool<Entity, Components>{dim}...
```
fixes the error.

skypjack:
Push upstream and added `-Wconversion` to the `CMakeLists.txt`. From now on, travis will tell us what's wrong in terms of conversions. ;-)

w1th0utnam3:
Cool. Yes, everything seems fine now.

skypjack:
Thank you very much. I'm closing this issue and considering it _solved_.

PR #4, opened 2017-05-27
Remove final specifier from all class templates

w1th0utnam3:
All class templates from the ECS are defined as `final`. In my opinion this is not sensible. In my concrete application I want to be able to forward declare the ECS class without retyping all template parameters (forward declaration of typedefs is not possible). The easiest way to do this would be to inherit from the EnTT types:
```
class EntityComponentSystem : public entt::StandardRegistry<EntityType, ...> {/* stays empty */};
```
now I can forward declare the `EntityComponentSystem` class in other parts of the project.

In general I don't see an advantage of using `final` in this context and I would propose to remove the  specifiers from the headers.

### DIFF
--- a/src/component_pool.hpp
+++ b/src/component_pool.hpp
@@ -20,7 +20,7 @@ class ComponentPool;
 
 
 template<typename Entity, typename Component>
-class ComponentPool<Entity, Component> final {
+class ComponentPool<Entity, Component> {
 public:
     using component_type = Component;
     using size_type = std::uint32_t;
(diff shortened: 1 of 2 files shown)


skypjack:
I can see the reasons to remove it from the registry and actually I can optimize a bit the component pool without the final keyword. Therefore we can safely drop them out. Anyway, there is no reason to remove it from the `View` classes. They are final on purpose and those classes are not meant to be extended by the users. Please, re-apply it back and I can accept the PR and make a few other changes. Thank you.

w1th0utnam3:
I added `final` to the `View` classes.

> I can optimize a bit the component pool without the final keyword

Can you elaborate on this? Do you mean e.g. making `private` members `protected` in order to properly subclass `ComponentPool`? Or are you talking about unrelated optimizations?

PR #46, opened 2018-02-26
Allow tests to compile on my box.

bjadamson:
I couldn't run the tests without this change. I'm not sure how it runs on your box without it, or more specifically what clang is doing that is different from whatever compiler you use.


skypjack:
It looks like [it's not a good one](https://travis-ci.org/skypjack/entt/builds/346529114?utm_source=github_status&utm_medium=notification)... I'm sorry. I cannot accept the PR as it is.

bjadamson:
I don't mind closing this, I'm happy to keep this change locally

skypjack:
Fair enough. I'll try to make it working during the next days, so that you don't have to keep it private and osx users can compile tests successfully. ;-)

skypjack:
@bjadamson Please, can you try branch `libcpp` **without** your change? I pushed a slightly different `CMakeLists.txt` that should work both on your mac and for the other users. This way you don't have to keep a private change locally. Let me know. Thank you.

bjadamson:
```libcpp``` compiles successfully on my box. However I'm using Linux mint, so I can't confirm is works on a mac.

skypjack:
Oh, ok, sorry, I thought you were using a mac because of libc++. Don't ask me why.
Instead, forgive me. I know it's annoying for a linux user to be called _mac user_. :-)
Debian Sid here, so I'm on your side. What is important is that it fixed the issue for which you created this PR.

I'm merging on `master` soon. Thank you.

PR #48, opened 2018-03-02
Idea

bjadamson:
Heyo, I ran into an issue where the copy constructor was being called on my component where I didn't expect.

I see that you thought about this case, and deliberately chose to use push_back because of POD's. Would you be opposed to allowing move-semantics for non-POD types?

Below I have an **idea**, if your willing to consider. It's by no means a final implementation or anything, but it should be close. Specifically defining of CXX17 would be worth discussing, and where to define it etc...


skypjack:
Hi, `std::is_pod` is deprecated and it will be removed with C++20. However I see what you mean and we can do it differently. You want to prefer emplace and thus direct initialization when possible instead of copy construction, right?

skypjack:
Wait a moment, I was still sleeping with the previous message!!

This:
```
instances.push_back(object_type{ std::forward<Args>(args)... });
```
It doesn't mean _copy_. I'm creating a temporary object of type `object_type`, thus the move constructor is picked up if viable.   
If you are observing a copy, that's probably because `object_type` isn't movable. `emplace_back` with a temporary object behaves exactly like `push_back` at the end of the day.

Can you copy here the type declaration?

What would solve the _issue_ in all cases is the use of `emplace_back` with the list of parameters, not with a temporary object.  
Anyway, note that `EnTT` requires types are movable. Otherwise you are going to experience compilation errors somewhere else sooner or later.

skypjack:
See [this commit](https://github.com/skypjack/entt/commit/7ca615a1c1bfe77942fa2c4b13f0a186f8b15a1d). It's a sfinae'd version of `construct` that uses in general `emplace_back` in all cases and favor `emplace_back` with arguments when possible.  
I'd suggest you to double check your type for the suspect is that it's not movable and it won't solve your issue anyway.

I'm closing the PR for I made the change you requested in a slightly different way and this no longer applies. Feel free to reopen it if required.

bjadamson:
I'll take a look. Thanks, I'm perfectly happy with my components being moved instead of copied. Actually that is what I was going for. This is a much cleaner looking implementation. I can't believe their deprecating is_pod in c++20... hard to keep up haha.

Feel free to leave this issue closed, thanks for addressing my concerns.

> However, no push_back in the sparse set anymore. ;-)

Fantastic. Houston mission accomplished.

skypjack:
Yep, [it's deprecated](https://stackoverflow.com/q/48225673/4987285). I'm not convinced of the reasons, but here it is. :-)

skypjack:
Yep, [it's deprecated](https://stackoverflow.com/q/48225673/4987285). I'm not convinced of the reasons, but here it is. :-)

Issue #56, opened 2018-03-12
Call for comments: blueprint/template


skypjack:
Another step towards the next version of `EnTT`.
The idea is to create a templating system (where template isn't intended as _C++ template_) with which to register predefined sets of components to assign to entities during creation.

Something along this line:
```
registry.blueprint<AComponent, AnotherComponent>("button");

// ...

auto entity = registry.create("button");
```

Doubts for which feedback would be appreciated:

* Should it be part of the registry and thus extend its API and introduce another vector internally or could it be an external tool?

* Runtime vs compile-time: `HashedString` can be used in constant expressions, otherwise they risk to introduce performance hits.
   In other terms, this has better performance at runtime:
   ```
   registry.blueprint<HashedString{"button"}, AComponent, AnotherComponent>();
   // ...
   auto entity = registry.create<HashedString {"button"}>();
   ```
   This one has a nicer API:
   ```
   registry.blueprint<AComponent, AnotherComponent>("button");
   // ...
   auto entity = registry.create("button");
   ```

* Is there any interest in such a feature? (Actually this one should have been the first question, btw)

In any case, I can't work on it for at least a week probably. Because of this, there is enough time to discuss the _best way_ to do it.

ArnCarveris:
@skypjack 
 
>Should it be part of the registry and thus extend its API and introduce another vector internally or could it be an external tool?

Prefer external tool.

>Runtime vs compile-time: HashedString can be used in constant expressions, otherwise they risk to introduce performance hits.
In other terms, this has better performance at runtime:

Important to have runtime creation.  make it configurable from data/file/resource.   
```
registry.blueprint<HashedString{"button"}, AComponent, AnotherComponent>();
// ...
auto entity = registry.create(HashedString {"button"});
```

skypjack:
@ArnCarveris I see your point. It makes sense.
Full support at runtime (instead of a mixed compile-time/runtime model) would make them usable also from within a plug-in system. Mod developers (if any) could write their own templates this way and then use them.
I didn't think about it actually. Interesting indeed.

ArnCarveris:
@skypjack 
>plug-in system. 

How to implement dynamic/shared library support? I mean custom components defined inside plug-in,

skypjack:
@ArnCarveris Actually I said plug-in but I meant script/runtime components in this case. My fault.
[...]

PR #74, opened 2018-05-05
Implement entt::Prototype

Kerndog73:
I had this implementation lying around for a while so I thought I should clean it up and open a pull request. Issue #56 has a discussion about this feature. Here's an example:

```C++
#include <entt/entity/prototype.hpp>

struct Position {
  int x;
  int y;
};
struct Sprites {
  int id;
  int frames;
};
enum class TileType {
  WALL,
  ENEMY
};

// not every entity can be initialized with EXACTLY the same components
uint32_t make(const entt::DefaultPrototype &proto, entt::DefaultRegistry &reg, Position pos) {
  uint32_t entity = proto(reg);
  reg.assign<Position>(entity, pos);
  return entity;
}

int main() {
  // imagine we’re initialising prototypes from files
  entt::DefaultPrototype wall;
  wall.assign<Sprites>(0, 4);
  wall.assign<TileType>(TileType::WALL);
  
  entt::DefaultPrototype enemy;
  enemy.assign<Sprites>(4, 4);
  enemy.assign<TileType>(TileType::ENEMY);
  
  // put some walls and enemies around the place
  entt::DefaultRegistry reg;
  make(wall, reg, {0, 0});
  make(wall, reg, {3, 3});
  make(enemy, reg, {3, 0});
  make(enemy, reg, {0, 3});

  return 0;
}
```

In your games, you would probably initialize the prototypes from JSON files instead of hardcoded values. If you're not defining parts of your game in a file then a prototype is essentially a factory function.

Thoughts?

### DIFF
--- a/src/entt/entity/prototype.hpp
+++ b/src/entt/entity/prototype.hpp
@@ -198,7 +198,7 @@ public:
     /**
      * @brief Assigns of replaces the given component
      *
-     * @tparam Comp Type of component to replace
+     * @tparam Component Type of component to replace
      * @tparam Args Types of arguments to use to construct the component
      * @param args Parameters to use to initialize the component
      * @return A refernce to the newly created component
--- a/src/entt/entity/prototype.hpp
+++ b/src/entt/entity/prototype.hpp
@@ -3,6 +3,21 @@
 
 #include "registry.hpp"
 
+namespace entt {
+
+/**
+ * @brief A prototype entity for creating new entities
+ *
+ * Prototype provides a similar interface to the registry except that Prototype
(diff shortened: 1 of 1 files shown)


skypjack:
Hi @Kerndog73 

It looks good, even though there are some things I would change (mainly to optimize in terms of memory usage instead of in terms of performance - I don't think prototypes are used usually along critical paths, am I wrong?).
Is it a problem for you if I _refine_ it somehow? I'm open to criticism once done. ;-)

Kerndog73:
It’s your library. I didn't expect you to merge straight away. Make whatever refinements you want. 😀

PR #87, opened 2018-05-31
Export CMake Config file and "modernize" CMake scripts

m-waka:
In modern CMake it's prefered to export a CMake configuation file for a package which can then be found by find_package(). This is also the prefered way to import external packages instead of find module scripts.
There is also no install target for entt, yet. Even for header only libraries one can use add_library as an INTERFACE library for generation of a target.

Some references for modern CMake are found in: https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1 and in Daniel Pfeifer’s C++Now 2017 talk Effective CMake (and cmake - Introduction and best practices, 2015).

So this PR tries to "modernize" the CMake scripts and exports a CMake configuration, which can be found by find_package(). It also adds an option for building the documentation, preventing an error when calling make install without having build the docs before.


CMakeLists.txt
```diff
@@ -4,6 +4,8 @@
 
 cmake_minimum_required(VERSION 3.2)
 
+include(GNUInstallDirs)
```
skypjack: `EnTT` fully supports also VS. Isn't this inherently wrong in that case?

CMakeLists.txt
```diff
@@ -36,44 +38,86 @@ message("*")
...
+# Add EnTT target
 #
 
-include_directories(${entt_SOURCE_DIR}/src)
+#message(STATUS "Compile features: ${CMAKE_CXX_COMPILE_FEATURES}")
+
+add_library(EnTT INTERFACE)
+
+target_include_directories(EnTT INTERFACE
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
```
skypjack: I know it's probably annoying, but currently `EnTT` uses 4 spaces for indentation. I'd prefer to keep it consistent throughout the whole project. Thank you.

CMakeLists.txt
m-waka: Actually here this only makes some variables avialable where only `CMAKE_INSTALL_INCLUDEDIR` is used and which evaluates to `include`. This should be correct or okay for most platforms. The other variables from `GNUInstallDirs` are not needed because the library is header only. We could also omit the include and replace the variable with a simple `include` if prefered.

skypjack: Don't worry, often I make comments just to better understand what's going on. If you confirm that it works also on VS, it's fine for me.

m-waka: I confirm it works under VS. Just tested it for playing safe.

m-waka: Ok, I just noticed that I used also the ${CMAKE_INSTALL_LIBDIR} variable from `GNUInstallDirs` for the exported and installed config file. This also works like a charm on Windows (VS) though I install it to one of the most common path from the unix conventions: `<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/`. But there seems to be different conventions where to install the config file on different platforms: https://cmake.org/cmake/help/v3.2/command/find_package.html

> CMake constructs a set of possible installation prefixes for the package. Under each prefix several directories are searched for a configuration file. The tables below show the directories searched. Each entry is meant for installation trees following Windows (W), UNIX (U), or Apple (A) conventions.

> This is merely a convention, so all (W) and (U) directories are still searched on all platforms.

skypjack: Is there a way to somehow _respect_ the conventions for the different platforms? `EnTT` has several users on Win after all.

m-waka: Yes. I'll try to adapt this to conform to the conventions dependent on the platform id using generator expressions.

skypjack: Good. Ping me if you find a solution. I'll leave the PR open while waiting for a feedback.  
Thank you for your efforts and your help to improve `EnTT`. Really appreciated.

m-waka: Okay generator expressions are not available in install for the destination so I had to use a variable `ENTT_INSTALL_CONFIGDIR`. The Windows convention for the location of the CMake config files should now be respected.

m-waka: Thank you, too. I would say it's ready to merge.

Just a little thing that I did not respect from the "modern CMake guidelines" is that they say one should "not abuse usage requirements: As an example, don’t add -Wall to the PUBLIC or INTERFACE section of target_compile_options, since it is not required to build depending targets." As this was already the case before, I did not change it.

skypjack: It's fine for me to keep on board `-Wall`. It helps quite a lot during development actually. The other way around that I could accept is to force `-Wall` only when compiling the tests. Do you think it's possible to do this?

m-waka: Yes you can add it to any target you want. So for the tests, should I add it only to the `odr` target or also to all the single test targets? By the way what stands odr for, One Definition Rule?

* Is there a reason you didn't add `-pedantic -Wall` for MSVC?
* I think the optimization and the libc++ options are also belonging to usage requirements. Do you want to keep them on board? - is libc++ improving performance?
* On my system (Archlinux with gcc 8.1.0 and cmake 3.11.2) the `-g` option is automatically appended when building in Debug mode so I get it twice. Also in Release mode it gives me the `-O3` optimization flag by default and I get it twice.

skypjack: >Is there a reason you didn't add -pedantic -Wall for MSVC?

If a reasons existed, I forgot it actually.

>I think the optimization and the libc++ options are also belonging to usage requirements. Do you want to keep them on board? - is libc++ improving performance?

If I remember it right, it was due to the fact that on mac it didn't compile without including libc++. Because of this I put a line to _prefer_ libc++ in general. We can rearrange it differently, not a problem. PR are compiled on the fly both on Travis and appveyor, so we know immediately if changes ruin something.

>On my system [...]

Well, I don't think we can assume this on all systems and putting twice those options isn't a problem, so...

---

In general, it looks like you are very skilled with cmake, so I tend to consider your suggestions.  
If you find we can further improve the build system of `EnTT`, feel free to continue pushing on the PR. We aren't in a hurry to merge it. ;-)

Milerius: I think for MSVC, we miss much things, especially as CMake Interface EnTT should use those flags on MSVC -> 
# /Zi - Produces a program database (PDB) that contains type information and symbolic debugging information for use with the debugger.
# /FS - Allows multiple cl.exe processes to write to the same .pdb file
# /DEBUG - Enable debug during linking
# /Od - Disables optimization
# /Ox - Full optimization
# /permissive- (c++ conformance)

m-waka: @skypjack Alright I think it's best to add the flags to the test targets then.
@Milerius I think these options/flags are no real requirements for using the library and so they don't have to be set as usage requirements on the EnTT target. But we can add some flags individually for the tests if needed. Projects depending on EnTT can define their own flags.

skypjack: @m-waka Fine for me. Do you plan to do it on your branch and therefore with this PR?

m-waka: @skypjack yes I will implement it on my branch with this PR tomorrow.

m-waka:
Alright, perhabs the added -pedantic and -Wall is the reason I will test and try to fix it on my Windows machine. So further commits will come.

m-waka:
So this builds now with MSVC2017 (at least on my system, let's wait for appveyor). Adding -Wall on MSVC was creating lots of warnings... I'm not sure whether these are problematic in any way.

skypjack:
I think appveyor isn't starting because a conflict within `googletest.in`.  
I just updated `GIT_TAG` to `master` because of #92.

m-waka:
So we have now two new CMake options which are by default `ON`:

option(ENTT_COMPILE_OPTIONS "Use compile options from EnTT." ON)
option(USE_LIBCPP "Use libc++ by adding -stdlib=libc++ flag if availbale." ON)

skypjack:
Really **THANK YOU** for your efforts!!
Let's see if travis and appveyor like this PR.

skypjack:
You did it!! Thank you!!
I'll review everything during the weekend and merge the PR.
Hope it's fine for you. Good job, welcome on board!! ;-)

m-waka:
Some tiny improvements/fixes, sorry...

skypjack:
Not a problem. Do you want to take your time to review it before to merge on `master`?  
If you need more time, we can leave open the PR for some days, as you prefer.

m-waka:
Okay, I will review it a bit and tell you if I think everything works and do some small commits if necessary.

m-waka:
I think this is fine now. We have now an additional build tree CMake config file generated besides the config file for the install tree.

skypjack:
Is it ignored by git or at least created within the build directory?

m-waka:
It ist here: `cmake/in/EnTTBuildConfig.cmake.in` and put into the buildtree by configure_file().

skypjack:
If it's fine for you, I'm merging it on a dedicated branch (`experimental`?), so as to be able to review it as a whole in a single commit before to port everything on `master`.  
I can change directly the PR in this sense, nothing to do your side.

m-waka:
As you want. :+1:  If there are any questions concerning the PR, feel free to ask.

skypjack:
@m-waka Do you have a copy of the branch? Because this is the first time I try to change the target branch on GH and I'm not sure what will happen!! :trollface:

skypjack:
So, what do we expect an user to do now to be able to easily use `EnTT`?  
I mean, if I were a new user that wants to use the library in a cmake based project, what should I do?

skypjack:
Another question. Why this:
```cpp
target_compile_features(
    EnTT
    INTERFACE cxx_std_14
)
```
Instead of this?
```cpp
set(CMAKE_CXX_STANDARD 14)
```

m-waka:
If it is already installed in your system you can easily call:
`find_package(EnTT 2.6 REQUIRED)
add_executable(myexec main.cpp)
target_link_libraries(myexec
    PRIVATE EnTT)
`
If it is not installed you can either clone and install it or add it as ExternalProject.
If you install it to a none default location (by setting the CMAKE_INSTALL_PREFIX variable to it) you have to add this location to the `CMAKE_PREFIX_PATH` variable to be able to find `EnTT`.

If you as developer of `EnTT` want to, you can also provide releases which include only the installed files (header files and install tree config files) or we can generate an installer with CPack.

You can still add it as subdirectory and as we also have a build tree config file it can then also be used without installing.

`target_link_libraries` is the way to express dependency relations. This way all the usage requirements (INTERFACE properties) of `EnTT` get imported by the target.
As I also provided the variable `ENTT_INCLUDE_DIRS` within both build tree and install tree config file you can still just call:
`target_include_directories(myexec PRIVATE ${ENTT_INCLUDE_DIRS})`
But this way you won't get the other usage requirements of `EnTT` imported.

**To your second question:**
Using `target_compile_features` for setting the C++ standard is the prefered way. This way you are expressing on a target basis the usage requirements of `EnTT`. These are then imported by dependent targets as explained above. You can even define the requirement on a feature basis e.g. `cxx_lambdas` and others.
On the other hand if you are setting the variable `CMAKE_CXX_STANDARD` and the project which is using `EnTT` alread sets this variable via a Toolchain file to e.g. C++17. This could probably have side effects. What you really wanted to express is that if you are using `EnTT` the importing target needs minimum flags for the C++14 standard.

I hope this was comprehensible.

skypjack:
Do you want to become the official maintainer of the `EnTT` build system, man? :1st_place_medal: 
It would be great to know that I can ask someone that knows what he's doing!! :+1:

skypjack:
Another question (forgive me, I'm trying to understand what you did in all the details).  
`EnTTBuildConfig.cmake.in` is auto generated. Shouldn't it be _git ignored_? I don't see why it's part of the new stuff.

m-waka:
Yeah, why not. Would be a great honor. I hope I will also improve my C++ knowledge while using `EnTT`.

`EnTTBuildConfig.cmake.in` is not auto generated. It gets configured by `configure_file` and is then put into the build tree. `configure_file` will replace `@PROJECT_VERSION@` and `@CMAKE_CURRENT_SOURCE_DIR@`.

skypjack:
In order to avoid polluting the build directory of external projects, shouldn't we replace `CMAKE_CURRENT_BINARY_DIR` and `CMAKE_CURRENT_SOURCE_DIR` with `entt_BINARY_DIR` and `entt_SOURCE_DIR`?  
In particular within `CMakeLists.txt` and `docs/CMakeLists.txt`.

m-waka:
As `CMAKE_CURRENT_BINARY_DIR` is the build directory that is currently being processed by cmake and `CMAKE_CURRENT_SOURCE_DIR` is the source directory that is currently being processed by cmake, this shouldn't pollute anything.

Issue #92, opened 2018-06-04
Compile error of `gtest` since Visual Studio 2017 v15.7.3


ArnCarveris:
After upgrading visual studio, while compiling any test emits following error:
```
\entt\deps\googletest\src\googletest\include\gtest\internal\gtest-port.h(996): error C2220: warning treated as error - no 'object' file generated
\entt\deps\googletest\src\googletest\include\gtest\internal\gtest-port.h(996): warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning.
```
That issue is know: [google/googletest#1373](https://github.com/google/googletest/issues/1373)

skypjack:
Yep. Can you help me to solve it?  
Find the file `googletest.in` (path `cmake/in`) and change the value of `GIT_TAG` to `master`, then try to compile.  
It should solve the problem. In this case, either open a PR or leave a message here and I'll update the project.  
Thank you.

skypjack:
Nevermind. Can you try branch `experimental`? Let me know if it works for you. Thanks.

ArnCarveris:
@skypjack fixed

skypjack:
Upstream.

PR #94, opened 2018-06-05
Adding parallel iteration to Raw, Persistent, and Single Component views.

vblanco20-1:
Implemented the functions par_each() for raw, persistent, and single component views. The syntax is the same as each(), but renamed to par_each(). Due to the overhead of std::parallel algorithms they are a clear pessimization on the benchmark suite. They have to be used with care and only in the cases where each iteration is actually expensive, as the parallel execution carries some serious overhead.

They are highly dangerous to use, as creating or destroying entities will instantly crash the application. Accesing shared resources from that its also very dangerous. Editing only the components for the iteration works completely fine.

The reason behind par_each() instead of the talked about each(std::execution::par) is that std::execution::par doesnt exist on non C++17 compilers, so it would complicate the design i think.

Parallel iterators are only in the C++ 17 standard, so the implementation is fully `#ifdef`-d to have it work on other configurations without breaking compiling. If the compiler is not C++17, it will just do the normal non-parallel version.

The current parallel execution is just completely basic std::execution::par, from std::algorithm, wich is not an optimal implementation. A optimized implementation that actually separates the iterator properly will run much better than this basic version.

Important detail. The current version of google test suite does not compile well with visual studio , with our without C++17. I had to fix some things on the test suite to be able to run the benchmarks from visual studio 17.

### DIFF
--- a/src/entt/config/config.h
+++ b/src/entt/config/config.h
@@ -7,4 +7,13 @@
 #endif
 
 
+//parallel algorithms only on C++ 17
+#if _HAS_CXX17
+
+#ifndef ENTT_HAS_PARALLEL_VIEW
+#define ENTT_HAS_PARALLEL_VIEW
+#endif
+
+#endif /* _HAS_CXX17 */
+
 #endif // ENTT_CONFIG_CONFIG_H
--- a/src/entt/entity/sparse_set.hpp
+++ b/src/entt/entity/sparse_set.hpp
@@ -64,12 +64,14 @@ class SparseSet<Entity> {
         using value_type = Entity;
         using pointer = const value_type *;
         using reference = value_type;
-        using iterator_category = std::input_iterator_tag;
+        using iterator_category = std::forward_iterator_tag;
 
         Iterator(pointer direct, std::size_t pos)
             : direct{direct}, pos{pos}
         {}
-
+		Iterator()
+			: direct{ nullptr }, pos{ 0 }
+		{}
         Iterator & operator++() ENTT_NOEXCEPT {
             return --pos, *this;
         }
@@ -507,11 +509,14 @@ class SparseSet<Entity, Type>: public SparseSet<Entity> {
         using value_type = std::conditional_t<Const, const Type, Type>;
         using pointer = value_type *;
         using reference = value_type &;
-        using iterator_category = std::input_iterator_tag;
+        using iterator_category = std::forward_iterator_tag;
 
         Iterator(pointer instances, std::size_t pos)
             : instances{instances}, pos{pos}
         {}
+		Iterator()
+			: instances{ nullptr }, pos{ 0 }
+		{}
 
         Iterator & operator++() ENTT_NOEXCEPT {
             return --pos, *this;
--- a/src/entt/entity/view.hpp
+++ b/src/entt/entity/view.hpp
@@ -1232,6 +1232,20 @@ public:
         });
     }
 
+	template<typename Func>
+	void par_each(Func func) const {
+#ifdef ENTT_HAS_PARALLEL_VIEW
+		std::for_each(std::execution::par,pool.view_type::cbegin(), pool.view_type::cend(), [&func, raw = pool.cbegin()](const auto entity) mutable {
+			func(entity, *(raw++));
+		});
+#else
+		std::for_each(pool.view_type::cbegin(), pool.view_type::cend(), [&func, raw = pool.cbegin()](const auto entity) mutable {
+			func(entity, *(raw++));
+		});
+#endif
+		
+	}
+
     /**
      * @brief Iterates entities and components and applies the given function
      * object to them.
@@ -1252,7 +1266,13 @@ public:
         const_cast<const View *>(this)->each([&func](const entity_type entity, const Component &component) {
             func(entity, const_cast<Component &>(component));
         });
-    }
+    }	
+	template<typename Func>
+	inline void par_each(Func func) {
+		const_cast<const View *>(this)->par_each([&func](const entity_type entity, const Component &component) {
+			func(entity, const_cast<Component &>(component));
+		});
+	}
 
 private:
     pool_type &pool;
--- a/src/entt/entity/view.hpp
+++ b/src/entt/entity/view.hpp
@@ -12,6 +12,9 @@
 #include "entt_traits.hpp"
 #include "sparse_set.hpp"
 
+#ifdef ENTT_HAS_PARALLEL_VIEW
+#include <execution>
+#endif
 
 namespace entt {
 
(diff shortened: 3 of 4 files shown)


vblanco20-1:
Added a 4th commit that adds a benchmark to compare the execution time of a system that does a "heavy" function. The heavy function is a very trivial prime checking, executed in a serial system and a parallell system to do a direct comparaison.
it shows a straight x8 speedup in my Ryzen CPU (8 cores).

vblanco20-1:
Made sure to have 3 different benchmarks for the 3 multithreaded variants. That way its easy to test if the 3 of them are multithreading properly.

skypjack:
> The current version of google test suite does not compile well with visual studio , with our without C++17.

I suspect it was due to #92 and probably it's worth it to try to run the benchmarks from visual studio 17 now.  
I'm a bit busy during this week, so it got me a while to update `master`.

skypjack:
If it's fine for you, I think I'll get the PR one piece at a time and merge it in the `cpp17` branch before to make it the `master` branch (see #95).  
I'll give you all the credits anyway, don't worry!! :-)  
I'm also trying to figure out what's missed to make the iterators actual forward iterators. It should be quite easy.

vblanco20-1:
@skypjack  do you think a Operator[] is possible? or a a "View.get<Component>(index)" function?. You could use other types of parallel for like #omp parallel for with that (should have lower overhead than the iterator based for_each parallel version

skypjack:
:thinking: Random access is somehow possible for persistent and raw views, absolutely. Single component views can support it as well. It's far more complicated in case of multi component views instead.

vblanco20-1:
I dont think that different views have different features is an issue. It would give options to the user. He could use a persistent view to iterate multiple components (faster, supports multithread, has the create/destroy perf hit), or normal view (slower, doesnt support multithread, doesnt have the perf hit).

skypjack:
I was thinking about this.  
Is `par_each` really necessary at the end of the day? I mean, if the iterators returned a the sparse set are forward iterators, the same are the ones returned by the views. In this case, wouldn't this be sufficient?

    std::for_each(std::execution::par{}, view.begin(), view.end(), [](auto &&...) { /* ... */ });

vblanco20-1:
How would the syntax of that work with multiple components? Will have a look and see how compares.

skypjack:
I think more or less the same as it works for the other views.  
Probably the example above is too short:

    std::for_each(std::execution::par{}, view.begin(), view.end(), [view](const auto entity) {
    auto &c1 = view.get<C1>(entity);
    auto &c2 = view.get<C2>(entity);
    // ... So on
});

skypjack:
With C++17, you can also do something like this:

    std::for_each(std::execution::par, view.begin(), view.end(), [&view](const auto entity) {
        auto &[c1, c2] = view.get<C1, C2>(entity);
        // ... So on
    });

In the meantime, I pushed on `master` all the changes required to turn iterators from sparse set in random access iterators and iterators from the multi component views in forward iterators.  
With the suggestion above, that is using `std::for_each` in combination with `view.begin()` and `view.end()`, it suffices to deal with multi threading probably.

I'm looking forward to your feedback before to decide how to proceed with the PR. Thank you.

vblanco20-1:
@skypjack  Absolutely massive improvement with the changes in master plus the "external" for_each parallel. The simulation now nearly maxes every one of the 16 threads of my ryzen CPU, and its nearly twice faster as my par_each version. Parallel overhead seems to be lowered by a huge amount, and systems now scale linearly.

image

skypjack:
Honestly, I made the changes a bit blindly and I didn't think to get that much! I'm glad to know it worked. :-)

The changes on `master` are partially taken from this PR, but I went further by defining iterators in the sparse set as random access iterators. Oh, I added also `operator[]` on the views because it was almost for free.  
Is there something else we can get from this PR? I mean, if the external `std::for_each` does the work, probably we can close the PR and get away with it?

As a side note, maybe it's worth it adding a dedicated section about multi threading with `EnTT` in the README.  
You proved that it's possible and quite easy to achieve indeed. Thank you very much!!

vblanco20-1:
@skypjack  the external for_each works great, i think its fine to not have par_each given that external for_each works fine. 

A multithreading section in the documentation, alongside an example of external for_each parallel would be great.  

Probably should add some multithreaded benchmarks to the test suite, you can use the ones i created there (with the prime check as "expensive" operation) to check the scaling, and a comparaison beetween the time taking with serial vs parallel versions would be great to check the overhead of parallelism. If you want, i can add those as a new PR.

The directx engine im building with EnTT is getting shape, Soon ill publish it open source. 

Feel free to close the PR if the discussion is finished.

skypjack:
>If you want, i can add those as a new PR.

I think it makes much sense to test it using the same laptop with which I made all the other tests. Thank you.  
Of course, I'll get your code and adapt it somehow. Thank you for the PR, it helped a lot actually!

>The directx engine im building with EnTT is getting shape, Soon ill publish it open source.

I'd be glad to add it to the list of projects in the README file. Feel free to ping me when you're done.

>Feel free to close the PR if the discussion is finished.

Sure. I'm adding some benchmarks and a few notes in the README file in the next days, then I'll close the PR. Thank you very much for your help to test it.

skypjack:
Damn it, what compiler are you using? It seems that neither GCC 7.3 nor Clang 6.0 are shipped with an implementation of the standard library that contains `<execution>`.

skypjack:
In the meantime, I added a few notes to the README file. Maybe I'll add some benchmarks after the switch to C++17.  
We can safely close the PR. Thank you again for your help. Really appreciated.

vblanco20-1:
@skypjack  im using Visual Studio 2017 latest patch. As far as i know, parallel algorithms are the absolute bleeding edge, so you probably need a latest version compiler that has good C++17 support (and enable C++17 in the build configuration)

skypjack:
Good to know. If you ever find room to improve `EnTT`, do not hesitate to write me!!

danielytics:
This is awesome! I like the performance gain, of course, but even more I like that it can use the external for_each and it also demonstrates that you can really use anything that accepts iterators (which means my plan of using Threading Building Blocks' tasks *should* also "just work").

skypjack:
@danielytics Well, the idea of making the views iterable was around since the very first commit. That's why they can be used in a range-for loop. However, I must admit that I did the minimum for them and this PR pointed it out. :-)

skypjack:
@vblanco20-1 

>With some modification to the Entt views, I was able to use “std::for_each()” on them, with the parallel execution policy to run the simulation lambda on every thread.

Did you do some other modifications to the view aside those we discussed together? As you can imagine, I'm pretty interested in the topic.

>I want to eventually modify it to use a Job system, and to be all in a fully parallel graph to reach 100% core usage over all threads.

If you come up with some changes or a new feature that can enrich further `EnTT`, do not hesitate to ping me. I'm always willing to work on the project to push it over its limits.

PR #97, opened 2018-06-09
fix warning on windows C4267

Milerius:
since size_t and unsigned int are different on windows we have to cast to entity_type.


skypjack:
Not exactly conforming to the rest of the codebase, but I'll merge the PR and adapt the changes. Not a great problem actually.  
Thank you for pointing out issues on Windows. Help on this is appreciated, for it's not the platform I use to work with `EnTT` on a daily basis.

PR #101, opened 2018-06-18
exposing empty delegate check

morbo84:
What do you think about it? This could be achieved using `delegate == {}` I guess, but maybe the `empty()` member function is more expressive...


skypjack:
Need some tests, but I can write them. I'm merging `experimental` on `master`, then I'll change the base of the PR and work on `experimental`, so as to be able to squash when done.

PR #107, opened 2018-06-30
Allow export of used Family template instantiation for shared libraries

DonKult:
Hi,

I know some header-only communities get edgy if you mention shared libraries, but please bear with me… so assuming you are crazy enough to write a shared library and some application(s)/game/whatever and want both to share their work and enTT between them you might end up in undefined behavior land due to symbol visibility as your (shared) components and co will end up having different identifiers in the two.

That is caused by the compiler (possibly) not seeing the template instantiations of the library and so instantiating them again for the application. This PR introduces a config flag to force the export of instantiations enTT makes so that they become public symbols of the library and are hence (hopefully) picked up by the application.

If that sounds complicated: Yes it is (in my eyes at least). Took me a while until I figured out the obvious… see the included GTest to see it in action, just flip the force flag and be dazzled. Works for me & tested on Debian only. I am all ears for alternative suggestions as there is probably some fun still if you have two libraries exporting these symbols, but I haven't tried that… that would be crazy, right? :wink:

So, and as a thanks for listening to a library zealot this PR includes ATM a second commit fixing a related README issue. Happy to split that out if needed, but feel free to just cherry-pick or whatever.


DonKult:
Correction: Don't flip the flag at home, just look at the output generated by the [clang-based tests](https://travis-ci.org/skypjack/entt/jobs/398637407#L250) as I seem to have failed and allowed only gcc to work… not sure yet why clang doesn't play along.

skypjack:
It looks like this PR cannot be merged as it is. It breaks tests on most of the platforms and thus it will probably break production software as well.  
Do you think you can look into the error returned by the CI system?

skypjack:
At a first glance, you're not exporting the right things. Types live in a namespace (the surrounding class in this case), you just ignored it. It seems that clang is right and GCC is wrong actually.

DonKult:
Yeah, my test was failing as it was depending on a [compiler-difference between gcc and clang]( https://stackoverflow.com/a/28827430). Marking all components in the library exported makes clang happy as well (without the fix clang at least handed out new identifiers as it shared the counter and not identifiers by a different counter as both compilers do without exporting the family symbols).

skypjack:
Your link confirms that GCC works even if it shouldn't. Right?  
Is there a way to fix this PR? As it is, it just breaks everything,. I don't think it would make much sense to merge it on `master`. Am I missing something?

Milerius:
Exuse me, but it seems that in my engine I use entt in plugins that are shared library and I have no problem currently.

DonKult:
@Milerius Are you building on Linux? On Linux all symbols are by default exported, but you can control this by setting `set(CMAKE_CXX_VISIBILITY_PRESET hidden)` globally or via `set_target_properties(TARGET PROPERTIES CXX_VISIBILITY_PRESET hidden)`. On Windows I think hidden is the default. Anyhow, libraries on any platform tend to us it to control which symbols become part of their public interface (as in which can be called by users) compared to what can be used only from inside the library. Handy for keeping ABI/API compatibility as well as to allow certain optimizations by compilers.

@skypjack I am not in the position of judging which compiler is right(er) here; compiler magicians^Wauthors shouldn't be second-guessed :wink:  Not sure what you mean breaking everything… Travis reports all green now?

skypjack:
What I don't like much are the changes to `config.h`. They ask for a maintainer and smells of incompleteness somehow.  
Isn't there an alternative way to obtain the same? As an example, a flag used to make family types public, so as to allow users to export them. This would move the `export` stuff on the other side of the border, that is where I prefer to put it so as to avoid problems. :-)

DonKult:
Regarding the potentially incomplete #ifdef: possible. Every project seems to use something slightly different. I see that cmake has also a command for it with `GENERATE_EXPORT_HEADER`, so that could potentially be used instead I guess.

Regarding moving it to the user: Could be done I suppose. If the user wraps the EnTT includes in a header of its own so (s)he can make the template instantiations before the EnTT headers can (s)he has practically the same effect. Just that every user needs to do it on his/her own and keep track of new instantiations while this would allow sharing the (probably small) workload and at least provide a hint for the befuddled if they run into this.

It is rather special-propose, so I can understand if you want to avoid this can of worms. Can you cherry-pick the README fix at least or should I open a PR just for that?

skypjack:
Not a problem to cherry-pick the changes to the README file.  
I know the others are for a specific purpose, but I usually try to find the best way so as to satisfy all the requests, if possible. Let's see if the _cmake way_ is feasible. Give me a couple of days to test it.

skypjack:
@Milerius Did you find the reason for which it compiles your side? As described by @DonKult you should have encountered problems as well, am I wrong?

skypjack:
@DonKult 

>there is probably some fun still if you have two libraries exporting these symbols, but I haven't tried that

I'm curious to know what happens in this case. I mean, if there exists a definitive solution that works also in this case, probably it's worth discussing it. Please, see #111 and continue analyzing the problem there. thank you.

Issue #109, opened 2018-07-02
Use Travis-CI to update packages


Kerndog73:
A few days ago, I learnt about continuous integration and setup travis with [one of my projects](https://github.com/Kerndog73/STELA) for the first time. CI is awesome! Test driven development is awesome! Why wasn't I doing this before?!

Anyway, I was thinking. Someone has to run a script or two when a new version of EnTT is released. I thought that maybe travis could do it. So I did a little research, and found this:

```yaml
deploy:
  provider: script
  script: bash update_packages.bash $TRAVIS_TAG
  on:
    tags: true
```

This will invoke `update_packages.bash` when a new tag is created. `update_packages.bash` could update packages for homebrew, vcpkg and any other package managers that might be supported in future like conan, spack, hunter, buckaroo, cget. The only change that will be required for the homebrew `update.bash` will be to clone instead of pull. I removed the section that does `brew audit` because it just produces annoying, nitpicky error messages.

This is just an idea and it might not be necessary but I wanted to see what you think.

This will avoid a repetition of the v2.7.0 incident. You ran the script, then I ran the script and created a merge commit. Dear oh dear 😅. That was my fault (sorry) but having Travis do the updating will mean that nobody has to worry about it ever again.

Milerius:
I'm totaly agree !

skypjack:
If we can bind a script only to tags, why not?  
Does the `tags: true` part serve this purpose?

Kerndog73:
@skypjack yep. https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

skypjack:
So, why did you open an issue instead of submitting a PR? :-)

Kerndog73:
Uuuuuuuuuum. Not sure actually. 😅

Kerndog73:
Maybe I'll do that now

skypjack:
:joy: I was joking, don't worry. However, if you can do it, it would be great!! Let me know. Otherwise I'll put a note in the TODO list for the next release.

Kerndog73:
I might as well. I have time.

PR #110, opened 2018-07-02
Update packages with Travis-CI

Kerndog73:
Turning issue #109 into a pull request.

This is difficult to test.


Kerndog73:
The script definitely works but I don't know of any way to test the whole thing until another tag is created.

Kerndog73:
Can the vcpkg package be updated with a script?

skypjack:
You can remove that part. It looks like [they have their own system](https://github.com/Microsoft/vcpkg/pull/3287) to updated registered packages.  
Is it possible to name script as `.sh` instead of `.bash` and add a `#!/bin/sh` line to the top?

Kerndog73:
I removed the `#!/bin/bash` part because the script will need executable permissions. So you'll have to do `chmod +x script.bash`. I think so

skypjack:
>I don't know of any way to test the whole thing until another tag is created

I like this risky stuff!! Isn't there a way to test it? Really? Otherwise we can create a v2.7.2 tag for this purpose. I don't think users will complain about it.

Kerndog73:
The script is easy to test but you can't really test the Travis part without creating a new tag and pushing to the remote repo.

skypjack:
>I removed the #!/bin/bash part because the script will need executable permissions. So you'll have to do chmod +x script.bash. I think

I didn't get it. What's the problem? We can just set the executable bit once and push the file, then everything should work as expected. Am I wrong? If I remember right, I did something with a project of mine (`uvw`) a lot of time ago.

Kerndog73:
I did some googling. It turns out that git tracks file permissions. Well you learn something everyday! I can't make this change now as I'm on mobile

skypjack:
I'm alone at home with my son (2 years old). I don't think I'l find the time to do that today. :smile: 
I can work on it for sure tomorrow afternoon. If you manage to do it in the meantime, well... that's good.  
Let me know. I'm willing to merge these changes and test them with the next tag.

Kerndog73:
The output from git suggested that the executable bit was set and the change was committed.

skypjack:
I made some small changes (ie `bin/bash` -> `bin/sh` so as to support different shells).  
Everything is made available on branch `issue_110`. You're still the author of the commit, of course.  
If you confirm that the resulting changes are somehow _valid_, I can merge everything on `master` and we'll test it with the next tag.

Kerndog73:
Looks fine to me. Let's see if it works!

Kerndog73:
The build errored! 😬

When `tags: true` is set, there is an implicit `condition: “$TRAVIS_BRANCH” = “master”`. When a tagged commit is pushed, `$TRAVIS_BRANCH` is set to `$TRAVIS_TAG`. This means that the implicit condition fails and we get `Already up to date!` in the build log.

The solution is quite simple.

```yaml
on:
  tags: true
  condition: “$TRAVIS_BRANCH” = “$TRAVIS_TAG”
```

skypjack:
Good point. Thank you. I'm updating `.travis.yml`.

PR #130, opened 2018-08-22
Port for VS 2015

eugeneko:
- Removed list initialization of static members;

- Replaced Identifier::get constexpr function with helper struct internal::IndexOf;

- Added default ctor for Monostate to let it be {}-initialized.

- Replace pivot trick in Registry::connect/disconnect with RemoveTypeFromTuple helper struct.

- Rename begin/end members in View and RealtimeView iterators in order to fix GTest trying to interpret iterators as containers.

- Add casts to Scheduler::then to fix function overload ambiguity.

### DIFF
--- a/src/entt/core/family.hpp
+++ b/src/entt/core/family.hpp
@@ -44,7 +44,7 @@ public:
 
 
 template<typename... Types>
-std::atomic<std::size_t> Family<Types...>::identifier{};
+std::atomic<std::size_t> Family<Types...>::identifier = {0}; // Workaround for VS2015: No list initialization
 
 
 }
--- a/src/entt/core/ident.hpp
+++ b/src/entt/core/ident.hpp
@@ -30,6 +30,18 @@ struct IsPartOf<Type, Current, Other...>: std::conditional_t<std::is_same<Type, 
 template<typename Type>
 struct IsPartOf<Type>: std::false_type {};
 
+template <class T, class Tuple>
+struct IndexOf;
+
+template <class T, class U, class... Types>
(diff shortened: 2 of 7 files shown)


skypjack:
Do you plan to update also the appveyoir config or I can proceed with the PR?

reworks:
the problem with vs 2015 is it doesn't have the greatest support for C++17.

skypjack:
@reworks 
Yeah, I think this is a temporary branch until VS2017 spreads out. I don't plan to port EnTT C++17 to VS2015 any time soon.  
Are you using the C++17 branch or the master one?

eugeneko:
@skypjack sorry, I’m totally infamiliar with appveyor.

@reworks And the problem with vs 2017 is that it’s ongoing experimental IDE and compiler that may break your code any time (this already happened two times), so I’d avoid using it for anything but experiments until it stabilized.

skypjack:
@eugeneko 
Not a problem, I'll configure appveyor myself.  
Just out of curiosity, are you an Urho user or a developer?

eugeneko:
@skypjack I'm user of Urho. I'm maintainer too. When I have enough motivation, I'm also developer...

skypjack:
@eugeneko So, when you plan to release a Urho version based on EnTT? :-)

reworks:
Are you sure your not using a beta or pre-release version? I must admit I've had issues with the occasional crash but I've never had a problem with any code written.

skypjack:
Merged into branch `vs2015`.

Issue #145, opened 2018-10-27
recursive error


pgruenbacher:
This library is what's first introducing me to c++17. I had to install gcc-8 g++-8 and set CMAKE_CXX_COMPILER to gcc-8 in the cmakeList, but otherwise I don't think that should have been an issue...

```
[  3%] Built target gtest
[  7%] Built target gmock
[ 10%] Built target gmock_main
[ 14%] Built target gtest_main
[ 16%] Built target odr
[ 20%] Built target dispatcher
[ 23%] Built target delegate
[ 25%] Building CXX object test/CMakeFiles/scheduler.dir/entt/process/scheduler.cpp.o
In file included from /home/pgruenbacher/Code/cpp/ecs/entt/test/entt/process/scheduler.cpp:3:
/home/pgruenbacher/Code/cpp/ecs/entt/src/entt/process/scheduler.hpp: In instantiation of ‘entt::scheduler<Delta>::continuation entt::scheduler<Delta>::continuation::then(Args&& ...) [with Proc = succeeded_process; Args = {}; Delta = int]’:
/home/pgruenbacher/Code/cpp/ecs/entt/test/entt/process/scheduler.cpp:78:38:   required from here
/home/pgruenbacher/Code/cpp/ecs/entt/src/entt/process/scheduler.hpp:65:21: error: cannot convert ‘entt::scheduler<int>::continuation’ to ‘entt::scheduler<int>::process_handler*’ in assignment
             handler = scheduler::then<Proc>(handler, std::forward<Args>(args)...);

```
And then I get inititely recursive error.

![screenshot from 2018-10-27 16-46-39](https://user-images.githubusercontent.com/3332598/47609186-ee279d80-da07-11e8-8ce4-ddc97ed7d778.png)

pgruenbacher:
and I am able to compile the library if I comment out handler = scheduler::then<Proc> line, and I think all tests will still pass except for the ./test/scheduler.

skypjack:
:thinking: 
Travis uses GCC8 as well as me on my laptop and I've never had this problem actually.
Moreover, that class is almost the same in the C++14 version. It's odd indeed.
What's the exact version of the compiler? I mean, 8.x.y

pgruenbacher:
this isn't a compiler issue that I can tell. I'm pretty sure you're supposed to be doing 
`handler = scheduler::then<Proc>(handler, std::forward<Args>(args)...).handler` in scheduler.hpp
[...]

PR #146, opened 2018-10-27
update the then methods to fix issue I was getting with g++-8 compiler

pgruenbacher:
see https://github.com/skypjack/entt/issues/145


skypjack:
Thanks. I've updated this. The `if` was useless. There is no chance a `continuation` object contains a null handler.

PR #166, opened 2018-12-08
added compile time version macros

prowolf:
Adds version macros to the version.h header.
ENTT_VERSION for the dot separated version
ENTT_VERSION_MAJOR for the major version
ENTT_VERSION_MINOR for the minor version
ENTT_VERSION_PATCH for the patch version


skypjack:
The main problem I see with this approach is that the version won't be available to users that don't rely on cmake to include `EnTT` in a project.

prowolf:
This approach puts the generated version.h file directly into the source tree. You could run cmake after changing the version number to update the file and commit it so it will be available to users that don’t rely on cmake.

skypjack:
Oh, I see. Good point. :+1:

prowolf:
Thank you for adding this!

skypjack:
Naaaa, thank you for the PR. Sorry if I'm late but I've been busy.  
If you can provide me with _<name surname aka nickname>_ string line, I can add you to the AUTHORS file as a contributor. :wink:

PR #167, opened 2018-12-11
UB/Crash in entt::insertion_sort

BenediktConze:
Currently, `insertion_sort::operator()` in core/algorithm.hpp is implemented like this:

    template<typename It, typename Compare = std::less<>>
    void operator()(It first, It last, Compare compare = Compare{}) const {
        auto it = first + 1;

        while(it != last) {
            auto value = *it;
            auto pre = it;

            while(pre != first && compare(value, *(pre-1))) {
                *pre = *(pre-1);
                --pre;
            }

            *pre = value;
            ++it;
        }
    }

Notice that `first` is incremented at the very beginning. This is undefined behaviour if `first` is the past-the-end-iterator. Therefore, the following example crashes with VC++ 14.16:

    #include <entt/core/algorithm.hpp>
    #include <vector>

    int main()
    {
        std::vector<int> v{};
        entt::insertion_sort sort{};

        sort(v.begin(), v.end())
    }

This can easily be fixed by comparing `first` and `last` at the beginning of the `operator()`-function (if `first` is past-the-end, `last` has to be past-the-end too).

### DIFF
--- a/src/entt/core/algorithm.hpp
+++ b/src/entt/core/algorithm.hpp
@@ -54,16 +54,16 @@ struct insertion_sort final {
      */
     template<typename It, typename Compare = std::less<>>
     void operator()(It first, It last, Compare compare = Compare{}) const {
-        if (first != last)
+        if(first != last)
         {
             auto it = first + 1;
 
-            while (it != last) {
+            while(it != last) {
                 auto value = *it;
                 auto pre = it;
 
-                while (pre != first && compare(value, *(pre - 1))) {
-                    *pre = *(pre - 1);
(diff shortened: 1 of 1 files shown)


skypjack:
Good point. Shame on me. Do you mind if I modify your commit to follow the coding style of the library?

BenediktConze:
No, not at all

skypjack:
Ok, thanks. I'll merge it on a separate branch so that I can add a test and squash everything in a single commit.
Thanks for the contribution. Do not forget to star the project if you like it.

PR #171, opened 2018-12-26
proposed swap method

pgruenbacher:


pgruenbacher:
since this involves swapping the std::unique_ptr<sparse_set<Entity>, which i don't think are accessible outside the registry. Maybe there's a way to validate that the two types are equivalent...
```
       // bool valid = std::is_same<C1, C2>::value;
        // assert(valid);
```
doesn't work.

there could maybe be a assert(sizeof(C1) == sizeof(C2))

pgruenbacher:
so I force the compiler to recognize the three Position components as separate types, but since they're implemented the exact same via inheritance, they should be fine.

pgruenbacher:
there should be an assert(ctype1 != ctype2) now that I think about it.

skypjack:
I see what you mean. However you're not swapping types, you're swapping pools, that are specializations of sparse sets and not guaranteed to follow the same rules of their components. What the registry casts under the hood are the pools, not the components, so it's risky.  
Btw probably we can obtain the same result working at a lower level, directly on components. The drawback is that performance wouldn't be the same as swapping two pointers unfortunately.

pgruenbacher:
yea it also needs to assert that the size of the two pools are the same too. 
but yea I'm using it in *production already and it seems to be working fine. Nice performance boost too and my system functions all got a lot simpler since they're just interacting with registry without worrying about double-buffer logic.

skypjack:
Let's try to find a slightly different approach that doesn't work out of an UB, so that we are guaranteed in the long term.

As a type safe alternative, wouldn't it be the same something like this (if `swap<T...>` existed)?
```
reg1.swap<T>(reg2);
```
This would work without falling in the dark corner of the UB and the performance are more or less the same of what you proposed.  
The sole drawback I see here is that consistency for entities isn't guaranteed out of the box, but things can be kept aligned probably with a couple of listeners set once and forever.

Kerndog73:
@pgruenbacher Swapping pool pointers is probably the fastest option as long as you know what you’re doing. You’ll have to keep the three pools in sync by adding and removing all three components together. Swapping pool pointers is a bit like calling `memcpy` or `reinterpret_cast`: fast if you know what you’re doing. This is a lot like writing to one `union` member and then reading from another.

We can do some compile-time checks for safety like `sizeof` should be equal and both should be `std::is_trivially_copiable_v`. Perhaps in debug builds we could check the two pools to make sure they are storing the same entities in the same place. I’m pretty sure that would just be a call to `std::equal`.

A safer solution with different performance characteristics (and no changes to `EnTT`) would be to have a `std::array<Menge::Math::Vector2, 3>` (that’s a long name for a 2D vector!) and then keep an index to the first position in a tag or something. Shifting is almost as fast as swapping pointers. `first = (first + 1) % 3`. You could keep all three indicies in tags so that you don’t have to add and modulus every time you need the second or third index.

Perhaps you could put your indicies in a `std::array<size_t, 3>` and `std::rotate` them. I suspect that would end up being just as fast as swapping pointers. Just a big list of `movq` instructions.

Iterating all three positions is a little faster now because they’re all packed together. However, iterating just one of the three is a little slower because 2/3 of the cache is full of data we don’t need.

The solution really depends on exactly how you are using these components. Swapping pools is faster if you’re iterating one set of positions at a time. Swapping array indicies is faster if you’re iterating all three sets of positions at a time.

pgruenbacher:
the reason I don't like having std::array<size_t, 3> components is because it means I lose certain flexibility. For instance, I'm tracking FSM state for my agents as a single component. Maybe somewhere down the future though, I decide I need to dual-buffer my FSM state, which can be likely for tracking N previous states. Now, all I need to do is add a FSM2 component, swap it, and viola everything still works. The only trick is to make sure a) I'm adding FSM and FSM2 components at same time (easy enough with prototype) and b) I need to make sure that fsm system is writing / reading to the two separate components. but all other systems that just read the FSM component don't need to be refactored. This allows for quick testing of buffering, better cache performance due to the packing of single components.

pgruenbacher:
the other option is to just expose the unique pointers and allow for user to do the swap manually, thus preventing misuse. But I think a good documentation and some user unit tests to ensure no corruption is plenty fine.

skypjack:
I thought of at least a couple of ways based on this PR to obtain the same result with a safer approach. I'll try to give you all the details soon, but I'm a bit busy during these days, so please be patient.

pgruenbacher:
sounds good.

skypjack:
Ok, sorry, I've been a bit busy during the last days. As I said, there exist at least a couple of solutions to achieve the same and get rid of the UB.  
Let's look directly to the problem: what you need is a way to create two pools of the same type (and swap them of course, but this is easy). A straightforward solution would be to provide a class template like this one:
```
template<typename T, auto I>
struct buffered_component {};
```
Then use some tricks with aliases and class templates to map it always on `component_pool<T>`, no matter what's the value of `I`. On the other side, `buffered_component<T, 0>` and `buffered_component<T, 1>` will have different families and therefore different pools we can safely swap.  
The drawback is that it's in charge to the user to keep the two pools in sync. to ensure that an entity has both the components, and so on. Is it bad? Not quite. You're in charge of it, but you've also the max flexibility.  
Note that the value of `I` is useless and serves only the purpose of differentiating the types. Btw you can do something like this with hashed strings: `buffered_component<my_type, "master"_hs>` and `buffered_component<my_type, "slave"_hs>`. I liked it much. :smile: 

The other solution around would be to use a similar component, but here `I` is an integer and is used to create N buffers (from 0 to N-1). Buffers are managed internally, when you add an entity and assign it one of the component, N-1 copies are assigned it. When you destroy an entity, all the copies are destroyed at once. In fact, it would be a dedicated pool aimed to treat transparently buffered components.  
More invasive a solution, less flexible too. Do we need more flexibility here? I don't know, you're the one with an use case, so I expect you have a reply on this. :smile:

pgruenbacher:
I suppose I like the idea of the swap api being `registry.swap<Component, 1, 2>` where internally the two types being swapped are `buffered_component<Component,1>` and `buffered_component<Component,2>`. Buffered components should always be part of the entity when first created I believe so managing the components should be on the user side. 

THe main thing I don't understand is if buffered_component is being used by the user too? Like do I do 
`using ReadPosition = buffered_component<Position, 1>` ?

skypjack:
>Buffered components should always be part of the entity when first created I believe so managing the components should be on the user side.

It's quite easy with the `dependency` utility. Set it and forget about it.

>do I use `buffered_component<Position, 1>` type as if it were a component? Or is it more like Position pos = registry.get<buffered_component<Position, 1>>(ent) ?

The latter. It's a tricky feature, so it requires more attention. But I see it could be useful.

pgruenbacher:
I see. I like the idea of registry allowing for pools of the same type so long as they are accessed by the buffered_component<T, int id>. This also eliminates me having to worry about making unique duplicates. Theoretically the ids can be used for non-buffering purposes too like so...
registry.get<Vector2, "position"_hs>
registry.get<Vector2, "orientation"_hs>

I mean I prefer explicit types for like above, but maybe someone would find that useful....
so instead of `buffered_component` I'd suggest something more generic like `uniq_component` or `duplicate_component`.

pgruenbacher:
much more succint.

skypjack:
We have a deal. :+1: :wink:

skypjack:
I just realized that persistent views break when you swap pools under certain circumstances. In fact, they won't work properly anymore if the two pools contain different sets of entities. There must be a way also to _reinitialize_ those views in this case.

skypjack:
Don't worry, I didn't forget your PR.  
I'm working on a new feature and I'm not sure yet the implementation on which we agreed will fit. Because of that, I'm delaying this a bit.

pgruenbacher:
shrug my fork works well enough for now.

skypjack:
Unfortunately, as expected, this doesn't fit well with groups (something I'm merging on `master` right now). I think the best would be to move double buffering client side probably.

skypjack:
Confirmed. This doesn't fit well with groups, at least as we discussed it.  
I'll add a note for buffered components in the TODO list and wait for the right idea.  
So far, I failed to find a proper solution that is both safe and doesn't require an amount of boilerplate hard to maintain. I'm sorry. :cry:

PR #174, opened 2019-01-13
allow for null entity members

pgruenbacher:
I ran into this when trying deserialize null entity member relationships.


pgruenbacher:
i didn't compile the test for master cause I'm too tired to update my cmake.
But the map() if null return null change in the header is what I'm using internaly now.

skypjack:
It makes sense actually. I think `entt::null` was introduced after the snapshot stuff and I forgot to fix this part. Good point.  
I'm merging it on a dedicated `branch` so as to make some changes before to merge on `master`.  
Thanks for pointing it out. :+1:

PR #176, opened 2019-01-24
Fix sparse_set invalid state when component constructor throws

corystegel:
# Preamble

First off, thank you for creating such an amazing piece of free software! 

I started creating a [similar library](https://github.com/cstegel/glomerate) to this when I surveyed the landscape of open source entity component systems a couple years ago and did not like what I found. I only needed it for one project at the time and then stopped working on it shortly after the project was done. When I searched for open source ECS recently I stumbled across `entt` and I am very impressed!

Any ways, on to the actual bug and fix...

# Bug Description

Previously, a component's constructor that throws would cause the sparse
set to think that the entity still exists in the set. This is because the underlying
sparse set that stores the entities will have its entry added before the component 
is added to the component set.

# Examples of Unintended Behaviour

This could cause a number of invalid memory access problems such as the following:

1) Exception triggers destructor of enclosing object that then tries to remove
   the component that was just added. SparseSet<entity_t, Component>::has() will
   return true for the entity but when destroy() is called "instances" will be empty
   so instances.back() will be invalid.

2) If the exception is handled then calling get(entity) for the same entity
   identifier that initially threw the exception will give a position for that
   entity even though it was not added. This can cause an invalid memory access
   or accessing the data of a different stored component.

### DIFF
--- a/src/entt/entity/sparse_set.hpp
+++ b/src/entt/entity/sparse_set.hpp
@@ -164,7 +164,7 @@ class sparse_set<Entity> {
         }
 
     private:
-        direct_type *direct;st
+        direct_type *direct;
         index_type index;
     };
 
--- a/src/entt/entity/sparse_set.hpp
+++ b/src/entt/entity/sparse_set.hpp
@@ -914,9 +914,9 @@ public:
      */
     template<typename... Args>
     object_type & construct([[maybe_unused]] const entity_type entity, [[maybe_unused]] Args &&... args) {
-        underlying_type::construct(entity);
 
         if constexpr(std::is_empty_v<object_type>) {
+            underlying_type::construct(entity);
             return instances;
         } else {
             if constexpr(std::is_aggregate_v<object_type>) {
@@ -924,7 +924,8 @@ public:
             } else {
                 instances.emplace_back(std::forward<Args>(args)...);
             }
-
+            // construct entity after component in case component constructor throws
+            underlying_type::construct(entity);
             return instances.back();
         }
     }
--- a/src/entt/entity/sparse_set.hpp
+++ b/src/entt/entity/sparse_set.hpp
@@ -164,7 +164,7 @@ class sparse_set<Entity> {
         }
 
     private:
-        direct_type *direct;
+        direct_type *direct;st
         index_type index;
     };
 
@@ -914,7 +914,6 @@ public:
      */
     template<typename... Args>
     object_type & construct([[maybe_unused]] const entity_type entity, [[maybe_unused]] Args &&... args) {
-
         if constexpr(std::is_empty_v<object_type>) {
             underlying_type::construct(entity);
             return instances;
@@ -924,7 +923,8 @@ public:
             } else {
                 instances.emplace_back(std::forward<Args>(args)...);
             }
-            // construct entity after component in case component constructor throws
+
+            // entity goes after component in case constructor throws
             underlying_type::construct(entity);
             return instances.back();
         }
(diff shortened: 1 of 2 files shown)


skypjack:
Hi @corystegel 
I'm glad you like `EnTT` and thanks for your contribution.  
What you say makes sense, even though I must admit I use always aggregate types that have no chance to throw exceptions. How did you spot it? Good job.

Would you mind if I edit your PR to follow the same style of the rest of the library? Just some aesthetic change, nothing substantial.

corystegel:
Thanks for the quick reply! Feel free to change any of the coding style (I tried to copy some of what I saw). 

I found this by running into the first situation I wrote about in the description. I'm currently using `entt` in a very atypical fashion for an ECS. I use a reference counting component that I attach to all entities to automatically delete them when all references fall out of scope. I had one component that would create and store multiple entities which each had their own components. One of these inner components threw an exception during a constructor (a mistake on my part). This eventually caused the root entity in this chain to lose all references and trigger `registry::destroy(entity)`. 

`sparse_set<entity_t, Component>::destroy()` gets called because `has(entity)` returns true. This function calls the following:

```c++
auto tmp = std::move(instances.back());
```
Since `instances` is empty so the move constructor access invalid memory.

skypjack:
>Feel free to change any of the coding style (I tried to copy some of what I saw).

Don't worry, it's almost fine indeed.  
I'm just a kind of _coding style nazi_, that's all. :smile:

I'll try to merge this on `master` in the evening.  
Thank you again for pointing it out. :+1:

OvermindDL1:
> I'm just a kind of coding style nazi, that's all. smile

Sounds like you want to setup clang-format with a custom configuration to run as a travis task to auto-format all code and PR's.  ^.^

corystegel:
> Sounds like you want to setup clang-format with a custom configuration to run as a travis task to auto-format all code and PR's. ^.^

I actually have my editor trigger `clang-format` whenever I save a file but I had to turn it off here and format manually since there was no `.clang-format` file. Without one, it was formatting the code with some default style. Sounds like it would be a nice addition to the project :)

skypjack:
Guys, you're right, definitely. However I've not a `.clang-format` set on my laptop. I must admit it.  
Is someone willing to decipher the current coding style and propose a proper config? :smile:

skypjack:
Uhm, the original PR failed to compile on appveyor... :thinking:

Kerndog73:
I’m currently running [unformat](https://github.com/johnmcfarlane/unformat) on the entt sources. This tool uses a genetic algorithm to find a clang format config that makes the smallest change to the sources. It’s been running for a few hours. At this rate, it would probably be faster to do it manually! BTW there should probably be a separate issue for this.

skypjack:
:smile: Did it finish in the meantime?

Kerndog73:
@skypjack There a few inconsistencies in the style so the tool couldn't find a good config. For example, you right-align pointers when used in variables and middle align pointers when used as return types. Clang-format doesn't provide granular control over pointer alignment. You just set it once and it's the same everywhere. The best config that the tool found would have had to change 1000 lines. Also, you need to set a column limit. Sometimes you wrap short lines into shorter lines. Sometimes you leave 120 column lines unwrapped. Set the column width in your IDE and stick to it.

Resolving these inconsistencies requires a decision from you. I don't really want to choose the style for you. You could start with one of the presets (LLVM, Google, Mozilla...) and then tweak it to your needs. There's probably and interactive website for fiddling around with styles. There's also a page that describes every option in the config with examples. [Here it is!](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

I'm sorry I couldn't be of more help. This is something that only the author can do. You're the one who's going to be looking at the code all day!

PR #195, opened 2019-03-01
Fix getting distance between forward iterators

WizardIke:
The documentation for a number of functions in registry stated that they take forward iterators but calculated the distance between iterators in a way that required random access iterators.


skypjack:
Good point. Thanks.

PR #217, opened 2019-04-07
rough proof of concept of swapping by using type member aliasing 

pgruenbacher:
this probably leaves options for swapping but also could enable components declaring their own container type. theoretically should work fine if the strong types are inheriting the container type. it's hacked right now so don't actually accept it, but you can use it as POC

### DIFF
--- a/src/entt/entity/sparse_set.hpp
+++ b/src/entt/entity/sparse_set.hpp
@@ -15,6 +15,40 @@
 #include "entt_traits.hpp"
 #include "entity.hpp"
 
+// https://stackoverflow.com/questions/11813940/possible-to-use-type-traits-sfinae-to-find-if-a-class-defines-a-member-type
+template<typename ...>
+struct void_type
+{
+     using type = void;
+};
+
+template<typename ...T>
+using void_t = typename void_type<T...>::type;
+
+#define HAS_TYPE(NAME) \
+template<typename, typename = void> \
+struct has_type_##NAME: std::false_type \
+{}; \
+template<typename T> \
+struct has_type_##NAME<T, void_t<typename T::NAME>>: std::true_type \
+{} \
+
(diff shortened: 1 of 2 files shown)


pgruenbacher:
I like to think that if this allows custom containers, then it can help enable reactivity by having containers that hole space for the old and updated values maybe...

skypjack:
I think I don't get what you mean with:
>by having containers that hole space for the old and updated values maybe

pgruenbacher:
> I think I don't get what you mean with:
> 
> > by having containers that hole space for the old and updated values maybe

idk really either. I just figured that if reactive system were implemented by EnTT it would need a custom container that holds both old and new values in memory.

skypjack:
Oh, well, not sure we have to maintain in memory both old and new values. It seems the widely accepted approach for that is to collect entities that changed and that's all, when you iterate them all what you can access is the current version of the component. If you really want both the values, you'll be still able to listen to the same event that fills the reactive system btw.

skypjack:
I'm closing this because swapping the pools this way would break groups and I can't find an easy way to work around it.  
We need a better approach to allow double buffering transparently, probably an in-pool automatism achieved with a specialization of sparse sets.  
That said, the idea of exposing the _actual type_ to use is smart and I'll try to elaborate in this direction to find a solution. :+1:

pgruenbacher:
swapping pools only breaks groups I think if there are groups that own one of the types, so perhaps one way would be to disallow groups owning types that are swappable. you can still partially own a swapped type though I think...

skypjack:
Yeah, but owning also _swappable_ types would be great for the performance, mainly because it's likely that those types are key ones (usually we speak of graphics stuff or so in this case). :wink:  
I think a solution like the one I want to experiment with for shared types could be applied also to dual buffering. More (hopefully) soon. :+1:

PR #218, opened 2019-04-08
A strong type alias wrapper

Kerndog73:
I thought about the problem that @iderik was having on Gitter. This might be a solution. It's a strong type alias wrapper. It acts like a `std::optional` that always contains a value. Here's a usage example:

```C++
struct vec3 {
  int x, y, z;
};

vec3 operator+(const vec3 &lhs, const vec3 &rhs) {
  return {lhs.x + rhs.x, lhs.y + rhs.y, lhs.z + rhs.z};
}

using position = entt::alias<vec3, struct position_tag>;
using velocity = entt::alias<vec3, struct velocity_tag>;
using rotation = entt::alias<vec3, struct rotation_tag>;

int main() {
  entt::registry reg;
  entt::entity e = reg.create();
  reg.assign<position>(e, 1, 2, 3); // works
  reg.assign<position>(e, vec3{1, 2, 3}); // also works
  
  // aggregate initialization works
  velocity vel{4, 5, 6};
  // access with with -> operator
  vel->x += 3;

  // must use static_cast here
  // this is good because you don't want to accidentally assign velocity to rotation!
  rotation rot = static_cast<rotation>(vel);
  
  // this is allowed though
  rotation rot1 = *vel;
  // but this isn't
  // rotation rot2 = vel;

  // you have to dereference to access overloaded operators
  // you don't have to do a cast here
  rot = *rot + *rot;
  // you don't have to do a cast here either
  rotation rot1 = *rot + *rot;
  // should we allow the above implicit conversions?
  // I think so

  // move-only types work too
  using ptr = entt::alias<std::unique_ptr<int>, struct ptr_tag>;
  
  ptr p;
  p = std::make_unique<int>(4);
  ptr q = std::make_unique<int>(8);
  p = std::move(q);
  // obviously not allowed
  // p = q;
}
```

The usage example speaks for itself. Looking at the code, I can't help but think I've just overengineered a solution to a problem that doesn't really exist. You could just do this:

```C++
struct position { vec3 p; };
struct velocity { vec3 v; };
struct rotation { vec3 r; };
```

This wrapper removes a few characters here and there and tries to be as transparent as possible. I don't know why I feel so inclined to use `constexpr` and `noexcept` so excessively when writing generic code. I never use `noexcept` and rarely use `constexpr` when writing "real" code! I haven't used `ENTT_NOEXCEPT` because it doesn't allow parametized `noexcept`. Should we add a `ENTT_NOEXCEPT_ARG` macro or remove the `noexcept` specifications?

This is still a POC at this point. I'm not sure if this is a good idea.

### DIFF
--- a/src/entt/core/alias.hpp
+++ b/src/entt/core/alias.hpp
@@ -1,0 +1,84 @@
+#ifndef ENTT_CORE_ALIAS_HPP
+#define ENTT_CORE_ALIAS_HPP
+
+#include <utility>
+#include <type_traits>
+
+namespace entt {
+
+template <typename Value, typename Tag>
+class alias {
+public:
+    using value_type = Value;
+    using tag_type = Tag;
+
+    template <typename... Args>
+    constexpr explicit alias(Args &&... args) noexcept(noexcept(Value{std::forward<Args>(args)...}))
+        : v{std::forward<Args>(args)...} {}
+
+    template <typename OtherTag>
+    constexpr explicit alias(const alias<Value, OtherTag> &other) noexcept(noexcept(Value{*other}))
+        : v{*other} {}
+    template <typename OtherTag>
+    constexpr explicit alias(alias<Value, OtherTag> &&other) noexcept(noexcept(Value{std::move(*other)}))
+        : v{std::move(*other)} {}
+
+    constexpr Value &operator*() & noexcept {
+        return v;
+    }
+    constexpr const Value &operator*() const & noexcept {
+        return v;
+    }
+    constexpr Value &&operator*() && noexcept {
+        return v;
+    }
+    constexpr const Value &&operator*() const && noexcept {
+        return v;
+    }
+
+    constexpr Value *operator->() noexcept {
+        return &v;
+    }
+    constexpr const Value *operator->() const noexcept {
+        return &v;
+    }
+
+    constexpr bool operator==(const alias &other) const noexcept(noexcept(v == other.v)) {
+        return v == other.v;
+    }
+    constexpr bool operator!=(const alias &other) const noexcept(noexcept(v != other.v)) {
+        return v != other.v;
+    }
+    constexpr bool operator<(const alias &other) const noexcept(noexcept(v < other.v)) {
+        return v < other.v;
+    }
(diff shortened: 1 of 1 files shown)


skypjack:
>Should we add a ENTT_NOEXCEPT_ARG macro or remove the noexcept specifications?

I wouldn't care much of the `noexcept`, the common use case is to disable exceptions anyway when use in real production code. :smile:  
Let's see what are the feedback from @iderik and if this worth it, I don't know if I would ever use such a tool but I don't see any drawback too. The problem is that the issue of @iderik seems due to an idiosyncrasy of `glm` and I'd rather make a PR/open issue there to solve it than try to fix it with a workaround.

iderik:
Thanks @Kerndog73 and @skypjack for taking your time!

If the dot operator could be overloaded (it cant be), the usage of entt::alias would have been looking so good.

I did open an issue over at glm (https://github.com/g-truc/glm/issues/891).

Milerius:
You probably want to look at https://github.com/doom/strong_type

Kerndog73:
@Milerius So you're suggesting we scrap `entt::alias` point people towards that library? I don't have a problem with that.

Milerius:
Some examples:

```cpp
using integer = st::type<
    int,
    struct integer_tag,
    st::arithmetic,
    st::addable_with<int>
>;

using acceleration = st::type<int, struct acceleration_tag>;

using speed = st::type<
    int,
    struct speed_tag,
    st::addable,
    st::subtractable,
    st::multiplicable_with<acceleration>
>;

using position = st::type<
    int,
    struct position_tag,
    st::addable_with<int>,
    st::subtractable_to<int>,
    st::orderable
>;
```

Milerius:
@kerndog 

I think that since the license is MIT we can totally do that yeah.

Milerius:
and we can do some magic alias:

using namespace entt = st; smth like dat

skypjack:
@Kerndog73 probably it makes sense not to add a small, almost incomplete class to maintain when there is a full-featured library out there to achieve the same purpose. I'm with @Milerius in this case.

Kerndog73:
@skypjack Yeah, I agree with you both.

PR #225, opened 2019-04-25
benchmark fix

eliasku:
only one entity has all the components


skypjack:
Good point. Thanks. :smile: :+1:

PR #227, opened 2019-04-26
exlude2test

pgruenbacher:
confirm that entt::exclude<> works for more than 1 component and that exlusions will work on both sets of components. also will replicate parentheses warning


pgruenbacher:
https://travis-ci.org/skypjack/entt/jobs/525174949#L838
here's replicating the warning

PR #228, opened 2019-04-27
Update README.md

ColinH:
A few minor changes to the wording...


skypjack:
Thanks for this. I spotted some errors but I can correct them on my own.  
I'll use this PR as a base point from which to start with a review of the README. :+1:

---

I'm out of home until Monday. I think I'll do it next week or so.

PR #229, opened 2019-04-27
Remove resource_handle class.

ColinH:
The `resource_handle` class adds no functionality to the `std::shared_ptr` it wraps, however it makes understanding the library more complicated by "obfuscating" said shared pointer.


Kerndog73:
`resource_handle` is a layer of abstraction. `std::shared_ptr` is an implementation detail.

Milerius:
Wtf is this pull request '_'

ColinH:
@Kerndog73 It just seems a superfluous abstraction since it doesn't add any functionality to the wrapped shared pointer. It also makes it harder to learn the library - everybody knows what a shared pointer is, now I need to remember that `resource_handle` is the same thing with a different name...

Milerius:
I'm using resource_handle and it's great

Kerndog73:
Don't think of `resource_handle` as a `std::shared_ptr`. Think of `resource_handle` as a flyweight _handle_ to a _resource_. What's hard to remember about that?

ColinH:
@Milerius Sure, it works and that's great, but is it _better_ than using a plain shared pointer? I'm just applying the KISS principle here.

skypjack:
I cannot accept this, I'm sorry.

---

I agree that nowadays a handle is more or less a narrower version of a shared pointer (eg you can't get explicitly the managed pointer with a `get` or similar). However, the point is another one and goes down to some future changes already on the roadmap.

The idea is to make this part allocator-aware (as well as the rest of the library actually) and therefore to get rid of shared pointers under the hood. As long as I use handles in my code and I stick with their tiny interface, I'll have to rewrite only my loaders. On the other hand, if I used shared pointers instead of handles, I'd have to touch the codebase of my software in several places and thus could be annoying (I use handles almost everywhere to manage resources).

Of course, this is based on the idea that allocators won't return shared pointers and that **is** definitely the case.  
It goes without saying that a cache that allows to manage memory with something smarter than a `new` is not even a topic to discuss. It's a must have that is to develop sooner or later.

Because of that, I prefer to continue using handles and be prepared for when they'll hide implementation details that are trickier than a shared pointer. Until then, please don't consider them only as _another name for a shared pointer_. It would be confusing, I agree.

skypjack:
I marked the PR as invalid.  
I'll leave it open for a few days to discuss further if needed. Then I'll close it if no one has a good reason to proceed otherwise. :+1:

ColinH:
If you are about to change the implementation in a way that keeps the current interface but gets rid of the shared pointers I agree that this PR doesn't make sense.

skypjack:
Indeed. I set up a few things that just _works_ for what I'm doing right now. However, the plan is to silently _evolve_ them when needed without great changes in my codebase. :+1:

ColinH:
Out of curiosity, why will you ditch the shared pointers when switching to allocators, is there a technical reason that it won't work, or won't you need a refcounting pointer anymore?

skypjack:
Mainly because shared pointers are a general purpose tool that works well in a lot of cases but isn't necessarily the best choice in this case.  
First of all, they expose an API I don't want to make available (eg I don't want to allow users to play directly with raw pointers to resources). Moreover, they rely on ref counting and atomic operations that waste cpu cycles for something you don't need most of the time in this context. Another thing that is harder to do with shared pointers and easier with custom handles is a lazy load of resources already referred by entities on a side registry or similar. And so on...
There are actually quite a few reasons for which a solution tailored to the specific problem can be better than a plain shared pointer.

ColinH:
Right, lazy load of resources won't work with a shared pointer as handle, gotya.

skypjack:
Yep. That's great because you can prepare your scenes in advance and load resources a bit at a time when needed, with hard deadlines in case.  
A shared pointer cannot allow it, indirection through handles supports it out-of-the-box instead. :+1:

PR #235, opened 2019-04-28
Peanuts

ColinH:
A few small things I found while browsing the source.

### DIFF
--- a/src/entt/entity/helper.hpp
+++ b/src/entt/entity/helper.hpp
@@ -44,7 +44,7 @@ private:
 
 
 /**
- * @brief Deduction guideline.
+ * @brief Deduction guides.
  *
  * It allows to deduce the constness of a registry directly from the instance
  * provided to the constructor.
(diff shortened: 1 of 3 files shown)


Milerius:
Const on integer are useless...

skypjack:
Why have you removed the `inline` somewhere and left it somewhere else?

ColinH:
I removed them from the file that I was mostly studying, but didn't want to go to more work before knowing whether you would even accept such a PR.

skypjack:
Well, actually I prefer multiple PRs for different requests.

As an example, the changes to `helper.hpp` and `delegate.hpp` are typos (_guidelines_ instead of _guides_) and there isn't much to say about them.  
On the other side, the ones to the sparse set file are in part _odds_ (eg `inline` is there for _<reasons>_ - last but not least also clarify the intentions to the future me - and I wouldn't get rid of it tbh) and in part meaningful. This kind of PRs are usually _cherrypicked_ line by line and merged for what is worth it.

Btw you can leave this open, not a problem. I'll pack the different changes in different commits fir my own. :+1:

ColinH:
Right, I can make separate PRs the next time.

PR #236, opened 2019-04-28
add single include of EnTT framework

Milerius:
Usefull for using in godbolt example:

https://godbolt.org/z/lqlKGS

### DIFF
--- a/scripts/amalgamate.py
+++ b/scripts/amalgamate.py
@@ -1,0 +1,299 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+# amalgamate.py - Amalgamate C source and header files.
+# Copyright (c) 2012, Erik Edlund <erik.edlund@32767.se>
+# 
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+# 
+#  * Redistributions of source code must retain the above copyright notice,
+#  this list of conditions and the following disclaimer.
+# 
+#  * Redistributions in binary form must reproduce the above copyright notice,
+#  this list of conditions and the following disclaimer in the documentation
+#  and/or other materials provided with the distribution.
+# 
+#  * Neither the name of Erik Edlund, nor the names of its contributors may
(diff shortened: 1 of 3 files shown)


Milerius:
if you want to use it yourself.

`python3 scripts/amalgamate.py -c scripts/config.json -s .`

skypjack:
This looks good. I'm guessing if we can automate it somehow using the CI or something similar.

Milerius:
You can now also comparing assembly generated by `EnTT` and possibly inprove the code this way

Milerius:
If you have python3 in the CI of course you can calling `python3 scripts/amalgamate.py -c scripts/config.json -s .`

Milerius:
Usually you will do it only for release i think

skypjack:
Yeah. The best thing would be if it triggered with a git hook on tags and created directly a PR or something on this line.

Kerndog73:
Another thing that’s along the same lines is a precompiled header. I’ve never used them before so I’m not sure how much of a build-time reduction they provide. Maybe cmake could compile the single include header into a precompiled header. Maybe distribute precompiled headers for major compilers for convenience.

I don’t know. Just a thought.

skypjack:
@Kerndog73 never used pch actually, I don't even know how to generate them, but I'm open to suggestion. :-)

Kerndog73:
@skypjack It doesn’t look like cmake has built-in support for PCH. We’ll have to use [cotire](https://github.com/sakra/cotire). Maybe it’s too much effort!

Maybe I should stop beginning my sentences with “maybe”.

Milerius:
In fact the single header is only for testing the framework out of the box on wandbor or godbolt for me

skypjack:
Where did you get this `.py` file to create an amalgamation? It seems to me it makes just a raw copy of everything. Isn't there something a bit more elaborate to reduce the size of the final file?

Milerius:
i stole it in nlohmann::json repository

PR #238, opened 2019-04-30
Type concat

pgruenbacher:
Is it actually useful to concat type lists? ... maybe yes if the group lists are big enough and the composition makes sense... if anything it could expose some more useful api opportunities. I'm at the c++ learning stage where complicated magical type folding logic *clearly* belongs in an api.

### DIFF
--- a/src/entt/core/type_traits.hpp
+++ b/src/entt/core/type_traits.hpp
@@ -13,10 +13,20 @@ namespace entt {
 template<typename... Type>
 struct type_list {};
 
+/* Recursive static method for confirming all types are unique */
+template <typename...>
+inline constexpr auto is_unique = std::true_type{};
+
+template <typename T, typename... Rest>
+inline constexpr auto is_unique<T, Rest...> = std::bool_constant<
+    (!std::is_same_v<T, Rest> && ...) && is_unique<Rest...>
+>{};
+
 /*! @brief A class to use to concat type lists. This is useful among other
 things for composing group inclusion and exlusion lists.*/
 template<typename... T1, typename... T2, template <typename...> class T>
 auto concat_types(T<T1...>, T<T2...>) {
+    static_assert(is_unique<T1..., T2...>);
     return T<T1..., T2...>{};
 }
 
--- a/src/entt/core/type_traits.hpp
+++ b/src/entt/core/type_traits.hpp
@@ -13,6 +13,12 @@ namespace entt {
 template<typename... Type>
 struct type_list {};
 
+/*! @brief A class to use to concat type lists. This is useful among other
+things for composing group inclusion and exlusion lists.*/
+template<typename... T1, typename... T2, template <typename...> class T>
+auto concat_types(T<T1...>, T<T2...>) {
+    return T<T1..., T2...>{};
+}
 
 /*! @brief Traits class used mainly to push things across boundaries. */
 template<typename>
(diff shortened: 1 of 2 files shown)


Milerius:
I'm against this naive implementation because it's horribly slowing compile time , for fast type list look at @doom implementation https://github.com/doom/meta/tree/master/include/meta/sequence/details

pgruenbacher:
well I mean it would only slow compile time where you specifically use it right? I imagine only using it in specific implementation files when I wanted to make big type lists to the api (for like component cloning) I get why the flatten_helper in doom/meta would be better, but it's so gross looking ha.

pgruenbacher:
by gross I just mean I'm just making fun of explicitly typing out all the possible combinations of template arguments, I'm not deriding doom who's code has helped me before...

it'd be cool if the type_cat could do variadic concatenating too actually...

skypjack:
Ok, this is the last PR to work on yet.  
If I got it, you are suggesting something like `std::tuple_cat` to concatenate type lists, right?  
What seems odd to me is the test for uniqueness. I see it fits your case, but I would rather offer it as a separate tool (eg to verify that a type list contains only unique types). There is no reason no to allow to merge a list with itself (granted, stupid example, but likely you got what I mean).

skypjack:
What about something along this line instead?
```
template<typename, typename>
struct type_list_concat;

template<typename... Type, typename... Other>
struct type_list_concat<type_list<Type...>, type_list<Other...>> {
    using type = type_list<Type..., Other...>;
};

template<typename Type, typename Other>
using type_list_concat_t = typename type_list_concat<Type, Other>::type;
```
Consider that we don't really need a function as it happens for `std::tuple_cat` because `type_list` doesn't contain actual values.

pgruenbacher:
so the usage would be something like 
```
using a = entt::type_list<A,B,C>;
using b = entt::type_list<D,E>;
const auto c = type_list_concat<a, b> 
```
?
The main use case would be something like registry.clone but with 10+ components

Kerndog73:
@pgruenbacher Not quite.

```C++
using a = entt::type_list<A,B,C>;
using b = entt::type_list<D,E>;
using c = entt::type_list_concat_t<a, b>; 
```

skypjack:
@pgruenbacher @Kerndog73 just a proposal, I'm open to suggestions. :+1:

Kerndog73:
@skypjack Somehow I feel like I've seen this before. Is it possible that `type_list_concat` was implemented and then deleted? Am I going mad?

Kerndog73:
@skypjack https://github.com/skypjack/entt/commit/70608df843b4c6f22893b484cc85cec0e3010b25#diff-df21289e1c62dc571a4e7d43d53c505e

I knew I'd seen it before!

skypjack:
@Kerndog73 good point, I completely forgot about it!! :smile:  
Actually, `type_list_cat` sounds better than `concat`.

pgruenbacher:
```
template<typename... Type, typename... Other, typename... List>
constexpr auto type_list_cat(type_list<Type...>, type_list<Other...>, List...) {
    return type_list_cat(type_list<Type..., Other...>{}, List{}...);
}
```
ah yes this is what i was wanting, just a variadic concatenater for type lists. Doesn't need to be used anywhere in the lib.... i just want it in entt closeby typelist as a reference

pgruenbacher:
saw the commit, cool, I'll test master this weekend.

PR #241, opened 2019-05-05
string view hashing

pgruenbacher:
I found myself needing to compare a hashed_string value against a string_view in an api call, so I thought it may be useful to have explicit handler for string views in entt:hashed_string.

```
    explicit constexpr hashed_string(const_wrapper wrapper) ENTT_NOEXCEPT
        : hash{helper(traits_type::offset, wrapper.str)}, str{wrapper.str}
    {}
```
kinda mystified...
for dynamic strings, how does hashed_string work when its constructor is constexpr?

### DIFF
--- a/src/entt/core/hashed_string.hpp
+++ b/src/entt/core/hashed_string.hpp
@@ -63,9 +63,24 @@ class hashed_string {
         const char *str;
     };
 
+    struct view_wrapper {
+        // explicit constructor on purpose
+        constexpr view_wrapper(const std::string_view curr) ENTT_NOEXCEPT: str{curr} {}
+        // constexpr const_wrapper(std::string_view curr) ENTT_NOEXCEPT: str{curr} {}
+        const std::string_view str;
+    };
+
     // Fowler–Noll–Vo hash function v. 1a - the good
     inline static constexpr ENTT_ID_TYPE helper(ENTT_ID_TYPE partial, const char *curr) ENTT_NOEXCEPT {
         return curr[0] == 0 ? partial : helper((partial^curr[0])*traits_type::prime, curr+1);
+    }
+
+    // string_view implemenation of hash
+    inline static constexpr ENTT_ID_TYPE helper(ENTT_ID_TYPE partial, std::string_view s) ENTT_NOEXCEPT {
+        for (uint i = 0; i < s.size(); ++i) {
+            partial = (partial^s[i])*traits_type::prime;
+        }
+        return partial;
     }
 
 public:
@@ -98,6 +113,10 @@ public:
      * @return The numeric representation of the string.
      */
     inline static hash_type to_value(const_wrapper wrapper) ENTT_NOEXCEPT {
+        return helper(traits_type::offset, wrapper.str);
+    }
+
+    inline static hash_type to_value(view_wrapper wrapper) ENTT_NOEXCEPT {
         return helper(traits_type::offset, wrapper.str);
     }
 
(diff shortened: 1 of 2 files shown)


skypjack:
Unfortunately, it cannot work this way. String views aren't null terminated. They are represented by a couple `const char *`/size.  
The idea is good, but the hashed string requires a set of changes for its own to support it.

You can either try to do it yourself or close this PR and open an issue for a feature request. :+1:

pgruenbacher:
mmmm... but what doesn't work? yes the instantiation doesn't work since the container is char* e.g. 
this is invalid-> `entt::hashed_string{string_view}` but the usage is just for hte hash function to work on the string view `entt::hashed_string::to_value(string_view)`

pgruenbacher:
essentially if I have an api that compares a string_view to a hash value in a component, i'd like to be able to  compare the hash value to the string view hash usin the same hash logic, which is why I've really only implemented the static to_value method. I think it makes sense that hashed_strings should only really be insantiated on constant string literals, not on dynamic string views.

btw I rush out these PR's cause i find it easier to discuss a PR than opening an issue if I have an idea of the code and want to dive into it.

skypjack:
It doesn't work as it is because currently the implementation looks for a null character in the string. String views don't have it necessarily, they have a size instead. It's not that difficult to implement what you're asking for, but this isn't the right way. That's all. :wink: 

Fine for me to keep open the PR. I'll try to implement it soon, it makes sense also for me this feature actually. :+1:

pgruenbacher:
but I already implemented the helper hash for the string view size() ? 
    
```inline static constexpr ENTT_ID_TYPE helper(ENTT_ID_TYPE partial, std::string_view s) ENTT_NOEXCEPT {
        for (uint i = 0; i < s.size(); ++i) {
            partial = (partial^s[i])*traits_type::prime;
        }
        return partial;
    }```

skypjack:
Oh... shame on me! I completely missed it from mobile. I'm sorry.  
Still, I'd rather modify the hashed string to accept a `const char *` and a size rather than directly an `std::string_view`. This way users can invoke it with a string view as `to_value(sv.data(), sv.size())`, but it will work fine also with old fashioned C-style strings. I can already imagine some uses for the latter case actually.  
I think it would be better in terms of API.

pgruenbacher:
yea you're right... I didn't realize string_view had data() method on it. still learning on it.

skypjack:
Exactly, I want to exploit the fact that it has a `data` member and a `size` member to add a more flexible version of `to_value` that works both with string views and dear old `const char *`.  
Something along [this](https://github.com/skypjack/entt/blob/experimental/src/entt/core/hashed_string.hpp#L110) line, that [can be used](https://github.com/skypjack/entt/blob/experimental/test/entt/core/hashed_string.cpp#L60) also with string views.  
Could it work?

Kerndog73:
How about an `std::string_view` overload as well?

skypjack:
@Kerndog73 that's the point of the discussion. I'd rather avoid to include another header just to offer an overload for a specific class when everything can just work out-of-the-box with a `to_value(const char *, std::size_t)`, that is a _raw-ish_ alternative that works not only with string views.

Kerndog73:
@skypjack Oh, I see. It currently includes `<cstddef>` and maybe `<atomic>` so it's very lightweight and you want to keep it that way.

skypjack:
Upstream. See https://github.com/skypjack/entt/commit/de737fc72decb4003d9e47cb6eb40869c6cfec58.

PR #248, opened 2019-05-21
Fix: remove garbage unicode U+FEFF failing compilation on Macos

mhammerc:
A faulty unicode character was present in `registry.hpp`.


skypjack:
Thank you very much! Feel free to update also the AUTHOR file and add your nick - mhammerc - if you want (lexical order). Otherwise I'll merge it as it is. :+1:

mhammerc:
It is done, thank you very much!

ronangaillard:
Same problem here, thx for fixing :)

Great job @mhammerc

PR #266, opened 2019-06-17
fix UB in insertion_sort

beppemarazzi:
Look at https://github.com/skypjack/entt/blob/e088f0f31ea5a55f71446e75b659017e8e123333/src/entt/core/algorithm.hpp#L64
because of post-decrement, when pre becomes equal to first, there is an UB (we are decrementing an iterator at the beginning). With MSVC compiler it is not a problem in release build (because we actually do not dereference the iterator), but in debug build an assert is raised from iterator post decrement operator (to reproduce it is sufficient to run tests in debug mode).
The only solution i found (without rewriting all) needs to use a `goto`... but probably is possible to rewrite it better!

### DIFF
--- a/src/entt/core/algorithm.hpp
+++ b/src/entt/core/algorithm.hpp
@@ -61,11 +61,16 @@ struct insertion_sort {
                 auto pre = it++;
                 auto value = *pre;
 
-                while(pre-- != first && compare(value, *pre)) {
+                while(compare(value, *--pre)) {
                     *(pre+1) = *pre;
+                    if (pre == first) {
+                        *pre = value;
+                        goto next;
+                    }
                 }
 
                 *(pre+1) = value;
+            next:;
             }
         }
     }


skypjack:
It's solved already on branch `experimental`.  
I made a promise in Milan, I didn't forget it!! :wink:  
I think I'll merge it on `master` later after dinner.

beppemarazzi:
Blame me!!! didn't checkout experimental branch.....

I found also DestroyByComponents test is asserting in debug. When the vector is empty it complains taking the address of the first element... probably it's an UB.
https://github.com/skypjack/entt/blob/e088f0f31ea5a55f71446e75b659017e8e123333/src/entt/entity/sparse_set.hpp#L141
My solution (but i think it's still an UB!!!) to workaround the assert is to replace with 
`return &direct->data()[pos];`

do you prefer a PR to track it???

skypjack:
I'm from mobile right now, so not sure about this but isn't that commit before the one that fixes the destroy by components?  
I mean, does it assert also with the latest commit from `experimental`?

beppemarazzi:
Mmmm... now I can't check. I left the pc in office... i'm going to try it
tomorrow morning. If it was already fixed, double sorry for the noise!!!!!

Il Lun 17 Giu 2019 21:05 Michele Caini <notifications@github.com> ha
scritto:

skypjack:
Nah, don't worry, thank you for pointing these out and for the help! :wink:

beppemarazzi:
Just to confirm that on experiment branch all is working.... Sorry again

PR #269, opened 2019-06-22
Conan: Support package in editable mode

Lawrencemm:
Add a method to the recipe that maps the include path to "src" when the package is put into "editable mode". See:
https://docs.conan.io/en/latest/developing_packages/editable_packages.html

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,5 +19,9 @@ class EnttConan(ConanFile):
         self.copy(pattern="LICENSE", dst="licenses")
         self.copy(pattern="*", dst="include", src="src", keep_path=True)
 
+    def package_info(self):
+        if not self.in_local_cache:
+            self.cpp_info.includedirs = ["src"]
+
     def package_id(self):
         self.info.header_only()


conanfile.py
```diff
@@ -19,5 +19,9 @@ def package(self):
         self.copy(pattern="LICENSE", dst="licenses")
         self.copy(pattern="*", dst="include", src="src", keep_path=True)
 
+    def package_info(self):
+        if not self.in_local_cache:
+            self.cpp_info.includedirs = ["src"]
```
skypjack: What does these lines do exactly? I saw it's an experimental feature in Conan that is likely to change in future.

Lawrencemm:
😅 May have got a little ahead of myself with this one but when you're
using conan you can say "instead of grabbing this package from a remote,
use my local working copy" except the local copy isn't "packaged" so the
headers aren't in the default "include" directory. This tells conan "when
I've said to use my local working copy instead of normal package, this is
where to find the headers" so that the generated build files point to the
right place.

skypjack:
Oh, ok, so the idea is that you put the sources under `src` and it uses the ones there instead of the ones from the conan's cache. Right?

PR #272, opened 2019-07-02
Support move-only parameters in delegate

Kerndog73:
I had a look at `std::function` and basically did what it does.

Here is the implementation of `std::function::operator()`.

```C++
template<class _Rp, class ..._ArgTypes>
_Rp
function<_Rp(_ArgTypes...)>::operator()(_ArgTypes... __arg) const
{
    if (__f_ == 0)
        __throw_bad_function_call();
    return (*__f_)(_VSTD::forward<_ArgTypes>(__arg)...);
}
```

`__f_` is declared as a `__base *` which has a call operator declared like this:

```C++
virtual _Rp operator()(_ArgTypes&& ...) = 0;
```

The changes were inspired by `std::function` but it was really just a whole lot of fiddling around until I got the tests to pass!

I made these changes on wip because I saw that you already made changes to delegate on wip and I didn't want to create a merge conflict.

### DIFF
--- a/src/entt/signal/delegate.hpp
+++ b/src/entt/signal/delegate.hpp
@@ -101,16 +101,16 @@ class delegate;
  */
 template<typename Ret, typename... Args>
 class delegate<Ret(Args...)> {
-    using proto_fn_type = Ret(const void *, std::tuple<Args...>);
+    using proto_fn_type = Ret(const void *, std::tuple<Args &&...>);
 
     template<auto Function, std::size_t... Index>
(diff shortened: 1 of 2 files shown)


skypjack:
I think I'll get most of your changes but I'll also add a test that uses an `std::unique_ptr` instead of defining a custom type. It's shorter. Why have you used a custom type here?

Kerndog73:
@skypjack I used a custom type because unit tests should be small and self contained. There’s nothing really wrong with using std::unique_ptr though

PR #284, opened 2019-08-09
how to assign values to change underlying data?

pgruenbacher:
WARNING, the second test fails. I'm afraid I dont understand the strategy for updating a primitive variable using meta without explicitly casting, which is a problem for strongly typed enums unless we know the enum type ahead of time. (can't cast to int). 

I do at least see the strategy for updating a member enum value using the meta_data.set()

### DIFF
--- a/test/entt/meta/meta.cpp
+++ b/test/entt/meta/meta.cpp
@@ -67,12 +67,19 @@ derived_type derived_factory(const base_type &, int value) {
     return {derived_type{}, value, 'c'};
 }
 
+enum class options {
+    optionA,
+    optionB,
+    optionC
+};
+
 struct data_type {
     int i{0};
     const int j{1};
     inline static int h{2};
     inline static const int k{3};
     empty_type empty{};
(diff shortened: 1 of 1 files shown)


skypjack:
Here I am. I think I don't get exactly this:
>// ok so how do we update the meta handle
>// with enum value (without expliclity using enum type casting)

Is your goal to modify directly a variable using the values for `options` registered through the meta type?  
In this case, probably you can do that by means of a couple setter/getter, but I must admit I've never had this requirement and I'll think if there exists a shorter way.

pgruenbacher:
ah your new test clarifies it for me. Thank you.
i may try extending the setter / getter to accept meta value as well.

skypjack:
Yeah. It's a bit convoluted but the fact is that meta types are all defined as if they were classes. Therefore we need to set a fake data member to set/get the value of a plain variable.  
If you come up with a good idea to simplify it, I'd be glad to discuss it. :+1:

PR #285, opened 2019-08-11
Prototype of LSD radix sorting algorithm and an example of working wi…

MrMartinIden:


skypjack:
I'm to merge this on `wip` and rework it a bit so as to make the algorithm _usable_ with `EnTT`. Just a note:
>sorts the elements in ascending order

Why do you explicitly mention this? Isn't it enough to pass reverse iterators to order everything in descending order?

MrMartinIden:
> Why do you explicitly mention this? Isn't it enough to pass reverse iterators to order everything in descending order?
Yes, you are right. I wrote a note because of my little experience working with iterators, and in fact I copied the algorithm taken from the link.

skypjack:
I've just realized that probably this cannot be used with a sparse set because of [this](https://github.com/skypjack/entt/blob/master/src/entt/entity/sparse_set.hpp#L528).  
Dumb question but have you ever tried to use this algorithm with the registry?

MrMartinIden:
Yes, I had to copy EnTT functions related to sorting, and slightly modify the body of the copied functions so that they could accept a Getter function object. In principle, I can throw off the necessary functions, the problem is that there are so few changes that they are almost similar to the original functions.

Issue #287, opened 2019-08-14
Bazel Support


zaucy:
Would the maintainers here be interested in supporting [Bazel](https://bazel.build/)? I have forked this repository here [bazelregistry/entt](https://github.com/bazelregistry/entt), but it would be great to have support on the official repository.

The fork I created is a work in progress and I wrote up several issues over there that I will eventually get to. If the consensus is positive I'd be willing to help in any bazel related issues that arise here in the future.

skypjack:
Hi @zaucy 

Thanks for your interest in the project.
I must admit I've never used Bazel but I've no reasons not to support it, especially if someone is willing to help with that. :wink:

Just a couple of questions:

* I guess the answer is _yes_ here. I'm asking just to be sure: is it possible to keep it in the project along with cmake and the others, so that it won't be a breaking change?

* Would you consider to close the issues currently open on your side before to merge it upstream?

I'd like it to be fully working before the merge.

zaucy:
> is it possible to keep it in the project along with cmake and the others, so that it won't be a breaking change?

Yep that is not a problem!

> Would you consider to close the issues currently open on your side before to merge it upstream?

Absolutely! I fully intend to fix the issues I have made on the fork before even making a pull request. ^^

skypjack:
So, why not! Go on. :wink:

Do you prefer to keep this issue open until it's done, so as to track your progresses?  
Otherwise feel free to close it and make a PR whenever you're ready.

I'm at your disposal for any question. :+1:

zaucy:
I'm okay either way. How about we close it once the PR (https://github.com/skypjack/entt/pull/291) has been merged :)

skypjack:
My only concern is: does it make sense at all to compile also tests with bazel?
So far, put aside the cmake build system (the original one for this project), all the other packaging tools or similar have been integrated so as to be able to include and use `EnTT`, never to compile tests.
Roughly speaking, this is because it would multiply for N the effort every time we add a test.

zaucy:
I'm not against removing them from my PR. In reality in order for someone to use entt with bazel they do not need the test BUILD files. I was only adding them in to have full _contributor_ support for bazel.

I can maintain a fork with the tests still in for myself since I've been migrating away from cmake in my own projects. That's no problem for me.

> Roughly speaking, this is because it would multiply for N the effort every time we add a test.

[...]

PR #290, opened 2019-08-18
benchmark to use entt::component

zaucy:
I ran into this while working on https://github.com/skypjack/entt/issues/287

PR #291, opened 2019-08-18
Bazel Support (#287)

zaucy:
closes https://github.com/skypjack/entt/issues/287

Results of `bazel test //...` with gcc 8.30:

Screenshot from 2019-08-17 19-43-07

Benchmark compiles/runs successfully (after merging https://github.com/skypjack/entt/pull/290) with `bazel run //test/benchmark`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,9 @@
+load("//bazel:copts.bzl", "entt_copts")
+
+cc_library(
+    name = "entt",
+    visibility = ["//visibility:public"],
+    strip_include_prefix = "src",
+    hdrs = glob(["src/**/*.h", "src/**/*.hpp"]),
+    copts = entt_copts,
+)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,41 @@
+workspace(name = "com_github_skypjack_entt")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "com_google_googletest",
+    strip_prefix = "googletest-release-1.8.1",
+    url = "https://github.com/google/googletest/archive/release-1.8.1.tar.gz",
+    sha256 = "9bf1fe5182a604b4135edc1a425ae356c9ad15e9b23f9f12a02e80184c3a249c",
+)
+
+http_archive(
+    name = "org_duktape",
+    url = "https://duktape.org/duktape-2.4.0.tar.xz",
+    strip_prefix = "duktape-2.4.0",
+    sha256 = "86a89307d1633b5cedb2c6e56dc86e92679fc34b05be551722d8cc69ab0771fc",
+    build_file_content = """
+cc_library(
+    name = "duktape",
+    visibility = ["//visibility:public"],
+    strip_include_prefix = "src",
+    hdrs = ["src/duktape.h", "src/duk_config.h"],
+    srcs = ["src/duktape.c", "src/duktape.h", "src/duk_config.h"],
+)
+    """,
+)
+
+http_archive(
+    name = "bazelregistry_cereal",
+    strip_prefix = "cereal-8629f40d932d57c5337d4557327f6f22436211b7",
+    url = "https://github.com/bazelregistry/cereal/archive/8629f40d932d57c5337d4557327f6f22436211b7.zip",
+    sha256 = "c983a7a2e16b153c3de022a0818d2f4836e510a3fc3bea9d3703de79f58a90a6",
+)
+
+# This is for bazelregistry_cereal
+http_archive(
+    name = "bazelregistry_rapidjson",
+    strip_prefix = "rapidjson-6b980984dacf689be8f65be823203b967c69da04",
+    url = "https://github.com/bazelregistry/rapidjson/archive/6b980984dacf689be8f65be823203b967c69da04.zip",
+    sha256 = "82187ba8de53bab3b4fb9e56d0bae81a96fa27a115e5a6ce14c70f0ef9338965",
+)
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ conan/test_package/build
 
 # IDEs
 .idea
+
+# Bazel
+/bazel-*
--- a/bazel/BUILD.bazel
+++ b/bazel/BUILD.bazel
@@ -1,0 +1,1 @@
+exports_files(["copts.bzl"], visibility = ["//:__subpackages__"])
--- a/bazel/copts.bzl
+++ b/bazel/copts.bzl
@@ -1,0 +1,10 @@
+_msvc_copts = ["/std:c++17"]
+_gcc_copts = ["-std=c++17"]
+
+entt_copts = select({
+    # Windows
+    "@bazel_tools//src/conditions:windows": _msvc_copts,
+    "@bazel_tools//src/conditions:windows_msvc": _msvc_copts,
+    "@bazel_tools//src/conditions:windows_msys": _msvc_copts,
+    "//conditions:default": _gcc_copts,
+})
--- a/test/benchmark/BUILD.bazel
+++ b/test/benchmark/BUILD.bazel
@@ -1,0 +1,11 @@
+load("//bazel:copts.bzl", "entt_copts")
+
+cc_binary(
+    name = "benchmark",
+    srcs = ["benchmark.cpp"],
+    copts = entt_copts,
+    deps = [
+        "//:entt",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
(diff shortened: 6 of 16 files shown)


WORKSPACE
```diff
@@ -0,0 +1,41 @@
+workspace(name = "com_github_skypjack_entt")
```
zaucy: This is the workspace name. Anyone who wants to consume an `EnTT` bazel package would prefix it with `@com_github_skypjack_entt//`. This is the standard naming convention. If there is an official EnTT website we could use that instead.

BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("//bazel:copts.bzl", "entt_copts")
+
+cc_library(
+    name = "entt",
```
zaucy: If someone where to use `EnTT` with bazel it would look something like this:

```python
cc_binary(
    name = "example",
    srcs = ["example.cc"],
    deps = ["@com_github_skypjack_entt//:entt"],
)
```

skypjack:
Looking at the results you posted, [this](https://github.com/bazelregistry/entt/issues/5) is to close as well, right? :-)

Thanks for the PR. I'll merge it upstream soon. :+1:

zaucy:
`bazel test test/entt/entity:helper` seems to be failing just on windows for me. I ran `ctest -C Debug` off of master on my machine and it was the same thing. I figured since it was failing with or without the bazel build files I would ignore it for now.

skypjack:
Can I ask what test fails? The CI doesn't return errors, so I don't know how to reproduce it.

zaucy:
This one: `test/entt/entity/helper.cpp`

Test log:
```
Running main() from gmock_main.cc
[==========] Running 5 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 4 tests from Helper
[ RUN      ] Helper.AsView
[       OK ] Helper.AsView (0 ms)
[ RUN      ] Helper.AsGroup
[       OK ] Helper.AsGroup (0 ms)
[ RUN      ] Helper.Dependency
test/entt/entity/helper.cpp(68): error: Value of: registry.has<double>(entity)
  Actual: true
Expected: false
[  FAILED  ] Helper.Dependency (0 ms)
[ RUN      ] Helper.Tag
[       OK ] Helper.Tag (0 ms)
[----------] 4 tests from Helper (0 ms total)

[----------] 1 test from Dependency
[ RUN      ] Dependency.MultipleListenersOnTheSameType
[       OK ] Dependency.MultipleListenersOnTheSameType (1 ms)
[----------] 1 test from Dependency (1 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 2 test cases ran. (2 ms total)
[  PASSED  ] 4 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Helper.Dependency

 1 FAILED TEST
```

zaucy:
Do you want me to make a separate issue for this? I'm able to reproduce it with the cmake build.

skypjack:
To be honest I've no idea on how to reproduce it. I've not a Windows machine available and appveyor doesn't return this error.  
Is there any chance you can help me to debug it on your machine?

Innokentiy-Alaytsev:
I've tried to reproduce the problem on current master (03949f7d60517a358e54e998d789bae8e5a914d6) and the test under mentioned above as failing passes. I've tried the following configurations:

 * Visual Studio 2017 (v141) Platform Toolset targeting Win32 and Win64

 * Visual Studio 2019 (v142) Platform Toolset targeting Win32 and Win64

Everything was built with Visual Studio 2019 16.2.1. I apologise for not providing the exact version of compilers - I have no idea about how to get it and don't have time right now for googling it.

P.S. Visual Studio 2019 (v142) Platform Toolset has problems with building meta test due to static assertion failure:

```
entt-src\src\entt\meta\factory.hpp(645,32): error C2607:  static assertion failed
entt-src\test\entt\meta\meta.cpp(181): message :  see reference to function template instantiation 'entt::meta_factory<data_type> entt::meta_factory<data_type>::data<& k,entt::as_is_t,std::pair<properties,int>>(const uint32_t,std::pair<properties,int> &&) noexcept' being compiled
entt-src\test\entt\meta\meta.cpp(183): message :  see reference to function template instantiation 'entt::meta_factory<data_type> entt::meta_factory<data_type>::data<& k,entt::as_is_t,std::pair<properties,int>>(const uint32_t,std::pair<properties,int> &&) noexcept' being compiled
```

zaucy:
> Is there any chance you can help me to debug it on your machine?

Yea I can help! Let me know what you want me to do.

These are the steps I do with cmake:
```
git clone https://github.com/skypjack/entt
cd entt/build
cmake -DBUILD_TESTING=ON ..
cmake --build .
ctest -C Debug
```

Update: The test passes when I build it optimized with bazel `bazel test -c opt test/entt/entity:helper`, but I haven't been able to get it to pass with cmake.

skypjack:
@Innokentiy-Alaytsev 

I'm updating the configuration for appveyor because I found they support also VS2019 since a few months ago.
It seems I can reproduce the static assert you hit.

@zaucy 

Let's see if I can reproduce also your error with appveyor. I've really no idea about what it's due to be honest.

skypjack:
@Innokentiy-Alaytsev 

The error with meta is due to some new amazing bugs of VS. :smile: 
I've put here and there some workaround that don't change the meaning of the code but make VS understand what's going on under the hood.

@zaucy 

Still investigating. The fact that clang and gcc as well as vs2017 work fine with it makes me think that it's an issue of vs2019, but I cannot really understand where it goes crazy.

skypjack:
@zaucy I saw that the `disconnect` doesn't work as expected in that test only. Unfortunately I've not a VS2019 to debug it, so some help in this direction would be appreciated!

Innokentiy-Alaytsev:
I'll try to investigate it deeper, but now I see strange behaviour with respect to lambdas in debug build.

With MSVS 2017 Platform Toolset (Win32 target) I see the following:



The important part: data pointers have the same value both for delegate prepared for removal and for the only connection in `calls`

With MSVS 2019 Platform Toolset (Win64 target) the picture is a bit different:



Here, the pointers of the delegate and the connection in `calls` are different! In Win32 configuration behaviour is the same.

Innokentiy-Alaytsev:
I've found a solution, which may grant me a title of Captain Obvious. I did not follow the EnTT code style because reasons and to improve my WTF-code's visibility (delegate.hpp:102 and later):

```c++
namespace fix {
	template<auto TFunction, class TRet, class... TArgs>
	struct Lambda {
		template <std::size_t... KIndex>
		static TRet Call (void const*, std::tuple<TArgs&& ...> args) {
			return TRet (
				std::invoke (TFunction, std::forward<std::tuple_element_t<KIndex, std::tuple<TArgs...>>> (std::get<KIndex> (args))...));
		}
	};
}
template<typename Ret, typename... Args>
class delegate<Ret(Args...)> {
    using proto_fn_type = Ret(const void *, std::tuple<Args &&...>);

    template<auto Function, std::size_t... Index>
    void connect(std::index_sequence<Index...>) ENTT_NOEXCEPT {
        static_assert(std::is_invocable_r_v<Ret, decltype(Function), std::tuple_element_t<Index, std::tuple<Args...>>...>);

        data = nullptr;

        fn = &fix::Lambda<Function, Ret, Args...>::Call<Index...>;
    }
// Remaining delegate class definition
```

I don't think that it is a good solution, but it works and all tests (except for meta, because I didn't pull fix) are green. Hope this will help finding a proper solution.

Innokentiy-Alaytsev:
I've just tested RelWithDebInfo configuration and the problematic test passes without the need for any fix.

skypjack:
@Innokentiy-Alaytsev 

I suspected it was something due to the decayed lambdas, you're confirming my guess. That's good.  
Does your fix solve the problem also with `RelWithDebInfo`?  
The way VS works is really obscure sometimes.

Innokentiy-Alaytsev:
@skypjack 
I'll check RelWithDegInfo configuration ASAP, but only after I get back home - I don't have MSVS2019 at work.

Innokentiy-Alaytsev:
BTW, different lambda types are seen in my screenshot for MSVS2019 despite me covering them with arrows. The types are the same in MSVS2017.

skypjack:
Yep but this looks a lot like something that is _wrong_ from the point of view of the language.
If I specialize a function template multiple time, there must be only one definition at the end of the day and therefore only one definition of the lambda inside it.  
I'll look if there exists an issue to VS for that.

skypjack:
@Innokentiy-Alaytsev I did something similar to what you suggested on `wip` (without the extra class but using `static` functions instead of lambdas). It still fails unfortunately.

PR #293, opened 2019-08-19
prototype iterator create

pgruenbacher:
i couldn't figure out how to deduce basic_registry<Entity>::size_type so I just defaulted to size_t out of laziness.

I'm ok with this not being merged but it was already in my head so wanted to show it at least as a feature api method when refactored to stomp.

### DIFF
--- a/src/entt/entity/prototype.hpp
+++ b/src/entt/entity/prototype.hpp
@@ -45,6 +45,7 @@ namespace entt {
 template<typename Entity>
 class basic_prototype {
     using basic_fn_type = void(const basic_prototype &, basic_registry<Entity> &, const Entity);
+    using grow_fn_type = void(basic_registry<Entity> &, size_t reserve);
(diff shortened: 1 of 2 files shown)


pgruenbacher:
the idea being to allow for batch creation without having to do registry.create<C1, C2, C3>(entities) which can be limiting.

pgruenbacher:
all it does is automatic reserve, a better way would be  to check if the component is default value or empty and if so then just do default constructible batch.

Kerndog73:
`batch_stomp`?

pgruenbacher:
yea maybe I should do batch_stomp PR instead.
I didn't realize that stomp tries to copy all components if none provided, which is great for clone and I suppose prototype logic too. I was worried I needed to declare the component types.
however this is probably a perf. issue for 100+ components + 1000 entities being stomped.

PR #294, opened 2019-08-20
batch stomp imp and test

pgruenbacher:
expand on stomp but iterate the range within the if conditional so that we're not repeatedly doing the run-time conditional check on each pool.

### DIFF
--- a/src/entt/entity/actor.hpp
+++ b/src/entt/entity/actor.hpp
@@ -10,7 +10,6 @@
 #include "entity.hpp"
 #include "fwd.hpp"
 
-
 namespace entt {
 
 
@@ -174,6 +173,7 @@ struct basic_actor {
         return entt;
     }
 
+
     /**
      * @brief Checks if an actor refers to a valid entity or not.
      * @return True if the actor refers to a valid entity, false otherwise.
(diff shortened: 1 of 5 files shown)


pgruenbacher:
added example prototype actor subclass for the specific purpose of having prototype actors that are easy to make copies of each other and stomp to others.

pgruenbacher:
the proto actor can be left out or integrated to basic_actor,

Kerndog73:
I think the management of prototypes is something that could be done on the user's side. `stomp` and `batch_stomp` make this really easy. Here's a rough sketch of something that a user might make. Prototypes are just `entt::entity`s. You could store them in a map or a struct or whatever.

```C++
class PrototypeManager {
public:
  entt::entity create(entt::registry &dst, ENTT_ID_TYPE id) {
    entt::entity e = dst.create();
    reg.stomp(prototypes[id], dst, e);
    return e;
  }
  void add(entt::entity e, ENTT_ID_TYPE id) {
    prototypes[id] = e;
  }
  entt::registry &registry() {
    return reg;
  }
  
private:
  std::unordered_map<ENTT_ID_TYPE, entt::entity> prototypes;
  entt::registry reg;
};

entt::entity loadEntityFromFile(entt::registry &, std::string);

PrototypeManager man;
man.add(loadEntityFromFile(man.registry(), "archer.json"), "archer"_hs);
man.add(loadEntityFromFile(man.registry(), "knight.json"), "knight"_hs);

entt::registry reg;
man.create(reg, "archer"_hs);
man.create(reg, "knight"_hs);
```

Using the registry API (instead of the prototype API or actor API) to manipulate entities means that you can more easily reuse code. The `loadEntityFromFile` function has no idea what a prototype or an actor is.

skypjack:
The best would be to find a way to use batch creation from pools. It's easy if the type is known, a bit harder otherwise because of the templated range.  
I'll try to find a way in this sense, it would be ways faster.

Kerndog73:
@skypjack Are generic iterators really necessary for this? How often do you find yourself filling a std::list with entities? Wouldn't a pointer and size be enough to cover the majority of use cases?

Kerndog73:
You could provide a generic iterator version that uses the pointer-and-size version to fill a std::vector and then copy to the iterator range.

skypjack:
@Kerndog73 You've views and groups, you can use their iterators with the function of the registries and that's pretty useful.

Kerndog73:
@skypjack Oh right, of course!

skypjack:
Probably, we can go on with a _best effort_ approach. When components are known, we can use the batch creation. When components are unknown, we can create one entity at a time through the erased function.

skypjack:
Batch `stomp` is available upstream. I've implemented it in a slightly different way than what you suggested with the PR, so I haven't merged this one. Can we close it?

PR #309, opened 2019-09-09
Mark CMake project explicitly as C++

janisozaur:
This prevents checking of C compiler

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ endif()
 # Project configuration
 #
 
-project(EnTT VERSION 3.2.0)
+project(EnTT VERSION 3.2.0 LANGUAGES CXX)
 
 include(GNUInstallDirs)
 


skypjack:
Good point, thanks. I'll merge it on `master` soon. :+1:

skypjack:
Unfortunately I must rollback this PR. If you enable `BUILD_MOD`, `duktape` is used to compile a test and therefore `C` is required. If you don't specify `LANGUAGES`, `C` and `CXX` are implicitly defined, that is all what is needed for `EnTT` and it's test suite.

Innokentiy-Alaytsev:
What about marking test using `duktape` as C and CXX language project?

skypjack:
@Innokentiy-Alaytsev Can we mark a single test? I didn't even know that.
It's fine for me. May I ask you to open another PR for that?

Innokentiy-Alaytsev:
I'll check it out. The first thing I thought of was "We can set the language in test project declaration", but now I see that there are no separate projects for tests. I'll check out two ideas:

 1. Create placeholder-ish project for all tests.

 2. Set languages for separate target.

Innokentiy-Alaytsev:
I'll prepare pull request later this evening. My idea is to try to call [`enable_language`](https://cmake.org/cmake/help/v3.0/command/enable_language.html) if `BUILD_MOD` is `true`.

skypjack:
Thank you. Really appreciated!! :wink:

skypjack:
Added to `wip` as you pointed out @Innokentiy-Alaytsev .
I put the call to `enable_language` in the file in the top directory as suggested by the documentation.

PR #310, opened 2019-09-09
Use Ninja on AppVeyor

janisozaur:
This brings up improved compilation speed and saner error messages.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,22 +23,34 @@ for:
       only:
         - image: Visual Studio 2019
     environment:
-      CMAKE_GENERATOR: "Visual Studio 16 2019"
+      CMAKE_GENERATOR: "Ninja"
+      VS_YEAR_VERSION: 2019
   -
     matrix:
       only:
         - image: Visual Studio 2017
     environment:
-      CMAKE_GENERATOR: "Visual Studio 15 2017"
+      CMAKE_GENERATOR: "Ninja"
+      VS_YEAR_VERSION: 2017
+
+install:
+  ############################################################################
+  # Install Ninja
+  ############################################################################
+  - set NINJA_URL="https://github.com/ninja-build/ninja/releases/download/v1.9.0/ninja-win.zip"
+  - appveyor DownloadFile %NINJA_URL% -FileName ninja.zip
+  - 7z x ninja.zip -oC:\projects\deps\ninja > nul
+  - set PATH=C:\projects\deps\ninja;%PATH%
+  - ninja --version
+
 
 before_build:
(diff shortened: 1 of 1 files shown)


skypjack:
How does this represent an improvement over the actual configuration? I don't get it to be honest.

OvermindDL1:
Ninja's primary use is significantly faster dependency tracking and job handling, it's often *quite* a bit faster on compiling larger projects than make/nmake/etc...  So I think it's just to reduce the build times.  In addition it's logs don't get mangled with errors from different threads if there is a failure to build like make/nmake do.

skypjack:
Does it use VC under the hood? Because the compatibility with MSVC is the whole point of using appveyor. I develop on linux usually, so the CI is what I use for this purpose.

OvermindDL1:
Yes, ninja is just the build orchestrator, it still uses gcc/clang/visual-studio-tools/icc/whatever.

janisozaur:
As was already stated, ninja is just a tool to invoke compiler. That and (almost) just that, which makes it fast, much faster than msbuild. You can verify what ninja does under the hood by passing `-v` option.
It also ensures each compiler invocation reports its output in single chunk, even from multiple instances.
From another ninja-ified project:

Before:
```
  LHVMViewer.cpp
C:\projects\openblack\src\LHScriptX\FeatureScriptCommands.cpp(240,24): warning C4244:  '=': conversion from 'double' to 'float', possible loss of data [C:\projects\openblack\_build_vs-16-2019-win64-cxx17_Debug\openblack.vcxproj]
C:\projects\openblack\src\LHVMViewer.cpp(171,54): warning C4267:  'argument': conversion from 'size_t' to 'int', possible loss of data [C:\projects\openblack\_build_vs-16-2019-win64-cxx17_Debug\openblack.vcxproj]
  MeshViewer.cpp
  Renderer.cpp
C:\projects\openblack\src\LHVMViewer.cpp(213,109): warning C4267:  'argument': conversion from 'size_t' to 'int', possible loss of data [C:\projects\openblack\_build_vs-16-2019-win64-cxx17_Debug\openblack.vcxproj]
```
After:
```
[32/50] Building CXX object CMakeFiles\openblack.dir\src\LHVMViewer.cpp.obj
..\src\LHVMViewer.cpp(171): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
..\src\LHVMViewer.cpp(213): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
```

Milerius:
Ninja is the best thing's that can happen in your life ;)

PR #319, opened 2019-09-16
Fix entt::collector.where

suVrik:
Hello! I think `entt::collector` has a bug. The `where` function actually changes the first added matcher, not the last one. In the following example both of `where` calls update the `group` matcher instead of preceding `replace` matcher.

```
entt::observer(registry, entt::collector.group<ComponentA, ComponentB>()
                                        .replace<ComponentA>().where<ComponentA, ComponentB>()
                                        .replace<ComponentB>().where<ComponentA, ComponentB>())
```

Here's the fix.

### DIFF
--- a/src/entt/entity/observer.hpp
+++ b/src/entt/entity/observer.hpp
@@ -83,7 +83,7 @@ struct basic_collector<matcher<matcher<type_list<Reject...>, type_list<Require..
     template<typename... AllOf, typename... NoneOf>
     static constexpr auto group(exclude_t<NoneOf...> = {}) ENTT_NOEXCEPT {
         using first = matcher<matcher<type_list<Reject...>, type_list<Require...>>, Rule...>;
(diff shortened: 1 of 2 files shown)


suVrik:
Btw, I personally would allow the `replace` matcher to be variadic as well.

skypjack:
Do you think you can write a test for this to add to `observer.cpp`?

>Btw, I personally would allow the replace matcher to be variadic as well.

The reason for which it is not is that the list is always intended as if the types were `and`-ed. It wouldn't be misleading for `replace` because:
* Intuitively, I read `replace<A,B>` as: _replace `A` **or** `B`.
* To support `or`-ed types we need to store aside some information, that is something I want to avoid.

suVrik:
Here's the test. I see the problem with variadic `replace` now.

skypjack:
Thank you for the test. I'll also add you to the list of contributors unless you prefer not to be there. Let me know.

suVrik:
That would be cool.

PR #322, opened 2019-09-18
Add namespace qualifier in CMake (close #321)

### LINKED ISSUE #321: Add an alias name for EnTT in CMake
I propose that you add a target alias for EnTT in CMake as to follow the convention of many other major libraries. Something like:

```cmake
add_library(EnTT INTERFACE)
add_library(EnTT::EnTT ALIAS EnTT)
```

This has numerous benefits according to the book `Professional CMake 2nd Edition by Craig Scott`. Some of which I'll quote here to back up why I think this would be beneficial:

**Quote from page 143:**
> Another` important aspect of names having a double-colon (::) is that CMake will always treat them as the name of an alias or imported target. Any attempt to use such a name for a different target type will result in an error.

Basically, without the alias name linking for `EnTT` in CMake without actually having the target available will make GCC attempt to use `-lEnTT` which can cause compile errors instead of generation time errors with CMake, that make it more clear something is wrong.

**Quote from page 144:**
> It is therefore more robust to link to namespaced names where they are available. Projects are strongly encouraged to define namespaced aliases at least for all targets that are intended to be installed/packaged. Such namespaced aliases can even be used within the project itself, not just other projects consuming it as a pre-built package or child project.

I'm not sure if including the quotes is allowed, but I do find them helpful to explain my point, and I will gladly remove them if it is inappropriate.

carlfindahl:
- Adds an alias to the library for in-source builds. IE: `add_subdirectory`
- Adds a namespace to the exported library so it also works with `find_package` and any other scenario where you don't build in-source

Tested on real world project wacman to both changes above and it built as expected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,7 @@ endif()
 #
 
 add_library(EnTT INTERFACE)
+add_library(EnTT::EnTT ALIAS EnTT)
 
 configure_file(${EnTT_SOURCE_DIR}/cmake/in/version.h.in ${EnTT_SOURCE_DIR}/src/entt/config/version.h @ONLY)
 
@@ -123,6 +124,7 @@ install(
     EXPORT EnTTTargets
     FILE EnTTTargets.cmake
     DESTINATION ${CUSTOM_INSTALL_CONFIGDIR}
+    NAMESPACE EnTT::
 )
 
 #


carlfindahl:
It seems to fail with a permissions issue with `apt`? I don't think that has anything to do with the changes made.

skypjack:
Nah, restarted, it doesn't depend on your PR. :+1:

carlfindahl:
You are welcome! Thanks for the effort in this library :grinning:

PR #325, opened 2019-09-25
This should allow any type to be used as Delta...

xissburg:
…beyond native numerical types.


skypjack:
Good point. `tick({})` should work as well.
Are you already in the list of contributors? :)

xissburg:
Ok I have changed to `tick({})`. This is my first contribution but I don't think I deserve to be part of that list just yet haha.

PR #326, opened 2019-09-27
Test process with both a complex data type and a fundamental type.

xissburg:


skypjack:
I chose a slightly different solution, so as to reduce the testing times. Thanks anyway for this PR from which I took inspiration and above all thanks for having indicated the problem with complex types!

xissburg:
Thanks! I'll have a look and see what I can learn from it.

PR #338, opened 2019-10-17
Properties list

Innokentiy-Alaytsev:
Implemented a function for adding a list of properties to meta-item.

### DIFF
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -730,7 +730,7 @@ class extended_meta_factory: public meta_factory<Type> {
 
 public:
     /**
-     * @brief Assigns properties to the last meta object created.
+     * @brief Assigns property to the last meta object created.
      *
      * Both the key and the value (if any) must be at least copy constructible.
      *
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -747,6 +747,24 @@ public:
         return *this;
     }
 
+    /**
+     * @brief Assigns properties to the last meta object created.
+     *
+     * Both the keys and the values (if any) must be at least copy constructible.
+     *
+     * @tparam Key Types of the properties keys.
+     * @tparam Value Optional types of the property values.
+     * @param pkey Property keys.
+     * @param pvalue Optional property values.
+     * @return A meta factory for the parent type.
+     */
+    template <typename... Property>
+    auto props(Property... ps) {
+		(unpack (0, ps), ...);
+
+        return *this;
+    }
+
 private:
     entt::internal::meta_prop_node **curr{nullptr};
 };
--- a/test/entt/meta/meta.cpp
+++ b/test/entt/meta/meta.cpp
@@ -20,7 +20,8 @@ Type get(Type &prop) {
 enum class properties {
     prop_int,
     prop_bool,
-    key_only
+    key_only,
+    prop_list
 };
 
 struct empty_type {
@@ -160,7 +161,12 @@ struct Meta: ::testing::Test {
                     .prop(properties::key_only)
                 .data<properties::key_only>("key_only"_hs)
                     .prop(properties::key_only)
-                .data<&set<properties>, &get<properties>>("value"_hs);
+                .data<&set<properties>, &get<properties>>("value"_hs)
+                .data<properties::prop_list>("prop_list"_hs)
+                    .props(
+                        std::pair{properties::prop_bool, true},
+                        std::pair{properties::prop_int, 0},
+                        properties::key_only);
 
         entt::meta<unsigned int>().data<0u>("min"_hs).data<100u>("max"_hs);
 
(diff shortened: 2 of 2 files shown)


skypjack:
I was to do the same in the afternoon but I had to leave and didn't push.
My question is: why a custom data structure? I'd rather use the old, good approach of _whatever supports `get<N>`_. This works out of the box with an `std::pair` as well as an `std::tuple` and an `std::array` or even an `std::variant`.
Isn't it better than a custom data structure I should use only for this purpose?

Innokentiy-Alaytsev:
I wanted those rvalue references. Not sure if it changes something. On the other hand `entt::a_prop` despite having an ugly name communicates the intention better than `std::pair`. I'll look into it.

skypjack:
Forget about data members having type `T &&`. It looks like an XY-problem. What's your goal exactly?

Innokentiy-Alaytsev:
Just want to pass the rvalues through, no other goal.

Innokentiy-Alaytsev:
The other reason is that it's simpler) I could not get anything meaningful in 10 minutes. The other thing is that I was curious how will parameter packs work when they have different length.

Innokentiy-Alaytsev:
Unfolding doesn't work for me with "anything with `std::get()`": https://godbolt.org/z/Dc3vhg

Using the recursive version is not fun because of compilation times.

skypjack:
>Just want to pass the rvalues through, no other goal.

Ok but... why? This way I cannot pass lvalue references anymore.

>Unfolding doesn't work for me with "anything with `std::get()`

Sure, [it does](https://godbolt.org/z/IREvud).

Innokentiy-Alaytsev:
> Ok but... why?

I guess it's just me being stupid) Right now I see no reasons for doing it this way.

> Sure, it does.

I thought about introducing an intermediate function the moment I turned my PC off after sending the previous comment.

There is a subtle point about "using anything compatible with `std::get()`" - sometimes it will do not what is intended:

 * An surrounding tuple is required if the user wants to add a key-only property that is of a type that is compatible with `std::get()` because otherwise, it may be unwrapped into a full property or into a property of a wrong type.

 * Trying to call `std::get()` on an `std::variant` instance that does not contain a value of the type with the provided index causes `std::bad_variant_access` to be thrown. Yes, it can be fixed during application testing but maybe unexpected for the users.

If you'd like I'll rework the PR to support "anything gettable" though. Of course, I shall remove those rvalue references in any case.

skypjack:
>[...] a key-only property that is of a type that is compatible with std::get() [...]

Ok, I won't take in consideration this corner case because it's unlikely you'll use something like that for the properties and a solution exists anyway if you did it.

>Trying to call std::get() on an std::variant instance that does not contain a value of the type with the provided index [...]

As above, I cannot even imagine why one should use an `std::variant` to set up properties to be honest but still a workaround exists for such an extreme case.

You can't stop all possible users from doing all the silly things that come to mind. As long as there is an alternative solution for them, move on and work on something else. :wink:

Innokentiy-Alaytsev:
What about a bargain: make the `properties()` function support anything "gettable" but have a simple convenience class `entt::a_prop` that is also gettable? Personally, I would really like not to have to write this class for every project or create a support library with this class alone. 

If you say "no" then I accept it and just make it work with anything "gettable" without any new DS's.

skypjack:
Ahaha... we don't have to make a bargain. I'm just trying to understand why you need it. :wink: 
I mean, why you prefer to rely on a custom defined class to do that rather than on widely used tools that are designed for the same purpose?
At the end of the day, you're just re-inventing a pair with that class. It just happens that the two members are named `key` and `value` rather than `first` and `second`. Isn't it? What am I missing?

Innokentiy-Alaytsev:
This small custom class better communicates the intent of creating a property. And the code is the same for both key-value- and key-only properties: `entt::a_prop{...}`. Having to use `std::tuple` for creating key-only properties, while all the other properties may be defined with `std::pair`, does not look beautiful to me. Of course, one may use `std::tuple` for everything but it only works if all properties are defined by the same person here and now. In legacy code with a team bigger than one person there is a chance of getting both `std::pair` and `std::tuple` (and maybe something else "gettable") in the same list. So it's not a problem of "it works" but of "how the code feels".

In fact, it's quite usual for me to overdo things because of this so some third opinion may be viable.

skypjack:
Oh, wait, there is a misunderstanding. I'm not saying I want to use **only** `std::get`-table objects. Key only properties shouldn't require to use an `std::tuple`, I agree. These are the forms I'm thinking of:
```
.prop(key1, value1)
.prop(key2)
.prop(std::make_pair(key3, value3), std::make_pair(key4, value4), key5, std::make_pair(key6, value6), ...)
.prop<&annotation1, &annotation2, ...>() // function based properties
```
What do you think about?

Innokentiy-Alaytsev:
I think this I can live with.

Innokentiy-Alaytsev:
The `unpack` function has made this PR kinda lame and trivial)

Innokentiy-Alaytsev:
A note regarding annotations. I just got a crazy idea: for enabling annotations (if they are used the way you've shown above) adding new properties passing a reference to the factory is required. With a reference for the factory, the annotations can do literally anything including adding new meta-members. That said, just sharing this crazy idea. Adding new meta-members from annotations will break the whole meta-state stack and all the following functions for adding properties will work with the wrong meta-item.

skypjack:
Ahahah... yeah, I think it would be like giving a loaded gun to users, more than C++ already is. :smile: 

>The unpack function has made this PR kinda lame and trivial)

I commented the issue about annotations to propose a _review_ that fits with the current implementation.
We can safely close this PR probably, let me know. :+1:

PR #347, opened 2019-10-27
Fixed build failure with exported types

Innokentiy-Alaytsev:
1. Changed the `entt::family::inner` template variable into a static function. The
   function does the same thing as the variable it's replaced - generates a new
   type id for the requiested type(s). The change is required to make MSVC happy
   when exporting/importing the families.

2. Changed `ENTT_DISPATCHER_EXPORT` so that only the type identifier generation
   function is exported. There is no actual need for exporting the stored type
   identifier because it will always be initialised by the same function that
   stores the identifier value as a static variable that is only initialised on
   the first function call.

The pull request may be merged if you decide it worths it. However, the main purpose is to run your GH actions and check if everything works.

### DIFF
--- a/TODO
+++ b/TODO
@@ -45,3 +45,4 @@
     * locator
   - update doc
   - update tests
+    * delete old tests (and files) with a_module, another_module, lib
--- a/src/entt/core/family.hpp
+++ b/src/entt/core/family.hpp
@@ -20,10 +20,6 @@ template<typename...>
 class family {
     inline static ENTT_MAYBE_ATOMIC(ENTT_ID_TYPE) identifier{};
 
-    template<typename...>
-    // clang (since version 9) started to complain if auto is used instead of ENTT_ID_TYPE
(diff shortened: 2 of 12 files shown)


skypjack:
It seems clang doesn't like this version though.

Innokentiy-Alaytsev:
I've broken it by force pushing the same version as on `skypjack/lib`

skypjack:
I'm not sure I get what you mean. Can I help you somehow?

Innokentiy-Alaytsev:
Don't worry. I've tried to see how your version of `lib` breaks GCC so i pushed it to my `wip`. The result is that PR is automatically closed) I'll try creating non-fork repo for EnTT to run Actions without the need for creating pool request.

skypjack:
It's not a problem for me if you want to create another PR and exploit `EnTT` to run GH Actions. :+1: :wink:

PR #351, opened 2019-10-29
Intoduced a macro for suppressing MSVC warning C4307

### LINKED ISSUE #348: Simplify proposed macro for hashed string warning suppression
The [current version](https://github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-C4307-integral-constant-overflow) of the macro for suppressing the warning regarding constant arithmetics overflow can be simplified to the following:
```c++
#if defined(_MSC_VER)
    #define HS(str)\
        __pragma(warning(suppress:4307))\
        entt::hashed_string{str}\
#else
    #define HS(str) entt::hashed_string{str}
#endif
```

This is not an important change but less code is better.

Innokentiy-Alaytsev:
The new macro can be used in user code for suppressing warnings when using
entt::hashed_string in constexpr context. The workaround provided in the FAQ

```c++
  #define HS(str)\
  __pragma(warning(push))\
  __pragma(warning(disable:4307))\
  entt::hashed_string{str}\
  __pragma(warning(pop))
  #define HS(str) entt::hashed_string{str}
```

can be reduced to

```c++
  #define HS(str) ENTT_HASHED_STRING_SUPPRESS_WARNING entt::hashed_string{str}
```

The new macro is also applied in the `ENTT_NAMED_TYPE` macro to reduce the
number of false warnings in user code. It is suggested that users should use
the macro also in their own specialisation of the `entt::named_type_traits`
template class.

The documentation was updated to mention a simplified form of warning
suppression and the newly added macro.

closes #348

### DIFF
--- a/docs/md/faq.md
+++ b/docs/md/faq.md
@@ -128,10 +128,8 @@ here is a workaround in the form of a macro:
 ```cpp
 #if defined(_MSC_VER)
     #define HS(str)\
-        __pragma(warning(push))\
-        __pragma(warning(disable:4307))\
+        __pragma(warning(suppress:4307))\
         entt::hashed_string{str}\
-        __pragma(warning(pop))
 #else
     #define HS(str) entt::hashed_string{str}
 #endif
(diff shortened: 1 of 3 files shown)


skypjack:
I'll merge this manually because it has conflicts with `wip`.

Innokentiy-Alaytsev:
It was not conflicting when I pushed it. From what I see only the TODO is conflicting so it's not something serious.

Innokentiy-Alaytsev:
I may resolve the conflict myself if you like

skypjack:
I don't want to waste your time. It's pretty small a change so who cares. Don't worry. :wink:

Innokentiy-Alaytsev:
We're in a bit of a pickle here because I don't want to waste your time 😄

skypjack:
I decided to merge only the changes to the file `faq.md` for two reasons:

* VS2019 no longer triggers the warning afaik.
* I'd rather avoid compiler specific directives like `#if defined(_MSC_VER)` for countless reasons (unless targeted to the boundary issue ofc).

Moreover, the same macro can be defined user side and therefore not having it as part of the library won't affect any functionality.

PR #354, opened 2019-11-03
[353] Tests for detecting the bug

Innokentiy-Alaytsev:
Four tests are added:

 * 2 tests for testing the good case: only one property is registered with
   either `prop` or `props` function call;

 * 2 tests for testing the bad cas: multiple properties are registered with
   either `prop` or `props` function call.

Signed-off-by: Innokentiy Alaytsev <alaitsev@gmail.com>

### DIFF
--- a/test/entt/meta/meta.cpp
+++ b/test/entt/meta/meta.cpp
@@ -2094,3 +2094,96 @@ TEST_F(Meta, Reset) {
     ASSERT_TRUE(entt::resolve("your_type"_hs).func("a_member_function"_hs));
     ASSERT_FALSE(entt::resolve("your_type"_hs).func("another_member_function"_hs));
 }
+
+
+TEST_F(Meta, ReRegistrationAfterResetSuccessSingular) {
+    struct singular_prop_ok {};
+
+    entt::meta<singular_prop_ok>()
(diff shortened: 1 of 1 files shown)


test/entt/meta/meta.cpp
```diff
@@ -2094,3 +2094,96 @@ TEST_F(Meta, Reset) {
...
+        .props("required_prop"_hs, "any_property"_hs);
+
+    ASSERT_TRUE(entt::resolve<plural_prop_fail>().prop("required_prop"_hs));
+    ASSERT_TRUE(entt::resolve<plural_prop_fail>().prop("any_property"_hs));
+
+    entt::meta<plural_prop_fail>().reset();
+
+    entt::meta<plural_prop_fail>()
+        .type("plural_prop_fail"_hs)
+        .prop("required_prop"_hs, "any_property"_hs);
```
skypjack: The error was trivial, I'm pushing the fix in a few minutes.
However, can you confirm this should be `props` rather than `prop`?

Innokentiy-Alaytsev: Yes, it should be `props`. It's most likely that I tested it with `prop` first and then introduced a test for `props` by copying and pasting and forgot to change the function name.

PR #358, opened 2019-11-09
Implemented scoped unregistering of meta-types

Innokentiy-Alaytsev:
Added an overload for `entt::meta` function accepting an `entt::meta_chain`
object reference for building a chain of locally registered types. The
`entt::meta_chain` instance will store an implicit list of types registered with
it and will call `entt::meta::reset()` for each of them upon its desctruction.

### DIFF
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -227,6 +227,20 @@ meta_any invoke([[maybe_unused]] meta_handle handle, meta_any *args, std::index_
 }
 
 
+struct meta_binding_node {
+    using unbind_function = void(*)();
+
+    meta_binding_node(unbind_function unbind) : do_unbind{unbind} {}
+
+    void unbind() {
+        if (do_unbind) {
+            do_unbind();
+        }
+    }
+
+    meta_binding_node* next = nullptr;
+    unbind_function do_unbind = nullptr;
+};
 }
 
 
@@ -238,6 +252,82 @@ meta_any invoke([[maybe_unused]] meta_handle handle, meta_any *args, std::index_
 
 template<typename, typename...>
 class extended_meta_factory;
+
+/**
+ * @brief A meta registration chain for automatic type unregistering
+ *
+ * A meta chain is an utility class for automatic type unregistering upon chain
+ * object lifetime end. Chains are non-copyable but movable so that meta-types
+ * registration life-time can be extended as long as required.
+ */
+struct meta_chain {
+    meta_chain() = default;
+
+    meta_chain (meta_chain const&) = delete;
+
+    meta_chain(meta_chain&& src) ENTT_NOEXCEPT {
+        *this = std::move(src);
+    };
+
+    meta_chain& operator=(meta_chain const&) = delete;
+
+    meta_chain &operator=(meta_chain && rhs) ENTT_NOEXCEPT {
+        ENTT_ASSERT(!binding_chain);
+
+        binding_chain = rhs.binding_chain;
+
+        rhs.binding_chain = nullptr;
+
+        return *this;
+    }
+
+    /**
+     * @internal
+     * @brief Appends a meta-binding node to the chain
+     *
+     * @param node A meta-binding node for appending to this chain.
+     *
+     * @note The attached node life-time must be such that it outlives the chain
+     * it's attached to.
+     */
+    void append(internal::meta_binding_node& node) ENTT_NOEXCEPT {
+        if (binding_chain) {
+            node.next = binding_chain;
+            binding_chain = &node;
+        }
+        else {
+            binding_chain = &node;
+        }
+    }
+
+    /**
+     * @brief Appends one chain to the other
+     *
+     * The appended chain will no longer unregister meta-types bound by it.
+     *
+     * @param chain A meta-binding chain for appending to this chain.
+     */
+    void append(meta_chain&& chain) ENTT_NOEXCEPT {
+        auto binding = binding_chain;
+
+        while (binding && binding->next) {
+            binding = binding->next;
+        }
+
+        binding->next = chain.binding_chain;
+
+        chain.binding_chain = nullptr;
+    }
+
+    ~meta_chain() {
+        for (auto binding = binding_chain; binding; binding = binding->next) {
+            binding->unbind();
+        }
+    }
+
+private:
+    internal::meta_binding_node* binding_chain = nullptr;
+};
 
 
 /**
@@ -813,6 +903,27 @@ private:
  */
 template<typename Type>
 inline meta_factory<Type> meta() ENTT_NOEXCEPT {
+    return meta_factory<Type>{};
+}
+
+
+/**
+ * @brief Utility function to use for scoped reflection.
+ *
+ * The same as `meta()` but also appends the registered type to the given
+ * meta-registration chain for automatic unregistration upon leaving the scope.
+ *
+ * @tparam Type Type to reflect.
+ * @param chain A meta-registration chain to which to append the newly registered type.
+ * @return A meta factory for the given type.
+ * @see meta
+ */
+template<typename Type>
+inline meta_factory<Type> meta(meta_chain& chain) ENTT_NOEXCEPT {
+    static auto binding_node = internal::meta_binding_node{[] { meta<Type>().reset(); }};
+
+    chain.append(binding_node);
+
     return meta_factory<Type>{};
 }
 
--- a/test/entt/meta/meta.cpp
+++ b/test/entt/meta/meta.cpp
@@ -2114,3 +2114,62 @@ TEST_F(Meta, ReRegistrationAfterReset) {
     ASSERT_TRUE(entt::resolve<props>().data("prop_bool"_hs).prop(props::prop_int));
     ASSERT_TRUE(entt::resolve<props>().data("prop_bool"_hs).prop(props::prop_value));
 }
+
+TEST_F(Meta, ScopedMetaUnbinding) {
+    struct first_scope_registered {
+        int data;
+    };
+    struct second_scope_registered {
+        int data;
+    };
+    struct third_scope_registered {
+        int data;
+    };
+
+    {
+        auto chain = entt::meta_chain{};
+
+        {
+            auto temp_chain = entt::meta_chain{};
+
+            entt::meta<first_scope_registered>(temp_chain)
+                .type("1"_hs)
+                .data<&first_scope_registered::data>("data"_hs);
+
+            chain = std::move (temp_chain);
+        }
+
+        {
+            auto nested_chain= entt::meta_chain{};
+
+            entt::meta<second_scope_registered>(nested_chain)
+                .type("2"_hs)
+                .data<&second_scope_registered::data>("data"_hs);
+            entt::meta<third_scope_registered>(nested_chain)
+                .type("3"_hs)
+                .data<&third_scope_registered::data>("data"_hs);
+
+            chain.append (std::move(nested_chain));
+        }
+
+        ASSERT_TRUE (entt::resolve<first_scope_registered>().data("data"_hs));
+        ASSERT_TRUE (entt::resolve<second_scope_registered>().data("data"_hs));
+        ASSERT_TRUE (entt::resolve<third_scope_registered>().data("data"_hs));
+
+        ASSERT_TRUE (entt::resolve("1"_hs));
+        ASSERT_TRUE (entt::resolve("2"_hs));
+        ASSERT_TRUE (entt::resolve("3"_hs));
+
+        ASSERT_TRUE (entt::resolve("1"_hs).data("data"_hs));
+        ASSERT_TRUE (entt::resolve("2"_hs).data("data"_hs));
+        ASSERT_TRUE (entt::resolve("3"_hs).data("data"_hs));
+    }
+
+    ASSERT_FALSE (entt::resolve<first_scope_registered>().data("data"_hs));
+    ASSERT_FALSE (entt::resolve<second_scope_registered>().data("data"_hs));
+    ASSERT_FALSE (entt::resolve<third_scope_registered>().data("data"_hs));
+
+    ASSERT_FALSE (entt::resolve("1"_hs));
+    ASSERT_FALSE (entt::resolve("2"_hs));
+    ASSERT_FALSE (entt::resolve("3"_hs));
+}


Innokentiy-Alaytsev:
There are some tests that fail. It's due to meta-types still being accessible through templated `entt::resolve` call. I guess it's expected behaviour. I think I'll rework tests.

Innokentiy-Alaytsev:
I've just seen that I didn't provide documentation. I'll do it soon.

skypjack:
Actually it's easier than this and we can exploit the already existing meta objects. There is no need to create a new meta object for this.
Long story short, we can add a field to `meta_type_node` to create a secondary implicit list aside `next`. This is a less obvious advantage: elements in the local and global implicit lists are ordered the same by construction. It means that removing an element is still linear even though you've to update two implicit lists.

Innokentiy-Alaytsev:
> Actually it's easier than this and we can exploit the already existing meta objects. There is no need to create a new meta object for this.

Do you mean `meta_binding_node` or `meta_chain` here?

skypjack:
Yep, exactly, we don't need any of them here. We can just reuse what we have already.

Innokentiy-Alaytsev:
But what will control the de-registration in this case? With the implementation above the `meta_chain` instance unregisters the types in the destructor. How do you plan to do it without additional objects? Through the destructor of the `meta_ctx`?

skypjack:
I'd add an `entt::reset()` function to destroy all the locally registered types where _locally_ means _on this side of the boundary_. That's RAII friendly in the sense that eg you've an object that in the constructor registers N types locally and de-registers all them on destructions. Ofc if you've two objects that register N types each, you've still to unregister them explicitly in their destructors but this is uncommon and I wouldn't care much of it. I'm much oriented to compromises that make common cases trivial, don't require many lines of code to maintain and still make possible to deal with corner cases.

skypjack:
Also, if I look at how I usually do things with meta, I've not a proper object that retains a state and therefore frees everything when it goes out of scope. Usually I've a free function like `setup` to register the types and a function `teardown` to unregister them. In this case, a _reset all_ function is much better than N invocation to `reset<T>`. The latter case is also definitely error-prone.

Innokentiy-Alaytsev:
I don't get how it will work. I have the following sequence of events in my mind:

1. The main executable is loaded. Some types are registered with its meta-context.
2. Some libraries are loaded. Each of them gets the main bind main meta-context and register some meta-types.
3. For some reason (e.g. plugin reload) one of the libraries decide to unregister all the types it had registered before so it calls `entt::reset()`.

If I understand the current implementation correctly, the sequence of events listed above will lead to **all** types being unregistered.

For things to work right storing a list of locally registered types is required. This means that the `local` meta-context should be used alongside the global one. Do I get your idea right?

skypjack:
More or less. We cannot use `local` because somewhere `local` is pointed by `global` (the main context) and you would clear all the meta objects if you decided to clear those on `local`. Therefore we must construct another list aside the main one but yes, the idea is that you described. :+1:

Innokentiy-Alaytsev:
Do you have any ideas on how to test the behind-boundary local types reset? I mean without creating a library.

Innokentiy-Alaytsev:
Don't worry, I've found the tests for it.

Innokentiy-Alaytsev:
Lazy me didn't find a way to do it with only a pointer to the local meta-type list - there is also a flag for local types.

Innokentiy-Alaytsev:
Well, the new version only works for MSVC due to the assumptions I've made. I'll keep working.

Innokentiy-Alaytsev:
And even under MSVC it doesn't work for static libraries...

Innokentiy-Alaytsev:
For now, I don't know how to solve this without introducing a separate context for registering local things.

skypjack:
I wrote something (still to finish though) that seems to work properly.
I've taken some parts of this PR and used a slightly different approach under the hood. The basic idea is that you don't want and don't need to break an eventually local const implicit list. All you want is to reset the meta types (that is detach their meta objects) and remove them from the context.
I should finish it tomorrow and share the code for comments. :+1:

skypjack:
See branch `wip`. I'm not satisfied with the naming and the idea of introducing a _generic_ meta factory. I'd rather use a free function similar to `resolve` to reset types but I'd like to avoid using the name `reset` because it's pretty generic and it doesn't give any information to the reader. Any idea?
That said, the implementation provided seems to work _as expected_. We can use it as a base for this feature.

skypjack:
Maybe `entt::unregister<T>()` and `entt::unregister()`?
As an alternative for `entt::meta<T>().reset()` and `entt::meta().reset();`.

Innokentiy-Alaytsev:
I couldn't understand how it works yet (I'll do it, gimme time) but I think there might be a bug. However, I may be wrong. Is the following right: the call to `entt::meta().reset()` should not unregister the types registered from libraries?

skypjack:
I was having a shower and thinking that we need a test for this actually. :smile:  
However, it shouldn't be the case because of [this](https://github.com/skypjack/entt/blob/wip/src/entt/meta/factory.hpp#L263). The _reset-all_ function doesn't iterate the global context but only a local chain.  
Am I missing something?

Innokentiy-Alaytsev:
I've reworked the `lib` test(s) for meta. I changed them in such a way that unregistration is performed gradually instead of all at once. Now they fail)

skypjack:
A good reason to keep it on `wip` until it's stable. :smile:

Innokentiy-Alaytsev:
BTW, the results and the overall mechanics of your implementation seems to be like those of my last attempt. Your version works differently but (if I understand correctly) has an additional meta-type list bound to the context alongside the main one. I didn't exactly get the purpose of the `hook`: it seems like it's set to some value, then its value is changed during `reset` and then it is never used again. If I understand right, you're using `meta_type_node::context` in approximatelly the same way as I used `meta_node::local` - for storing a *local* list of meta types. How far from the truth am I?

Innokentiy-Alaytsev:
I've changer `ASSERT_*` to `EXPECT_*` for now to make tests a bit more informative (but noisy). This change may be reverted after we fix everything.

skypjack:
Oh, yeah, well, tests fail because for `.so` we don't have the `-fvisibility=hidden` set to true (they pass on my laptop this way) and for `.a` we have no way to make visibility hidden, therefore there doesn't exits the concept of _locally created_ for a meta type.
Being the local context a static variable, it suffers from the opposite problem of what we observe on Windows. That is, the _local context_ is local to all the contexts, no matter what.

Probably it's just wrong the idea of deleting only locally created meta objects and we should always reset all the meta types present on the context. In case you want just to shutdown a plugin, you can still unregister objects on a per-type basis and all the cases are easily treated this way.

Innokentiy-Alaytsev:
My initial proposal for this feature doesn't have these problems, BTW.

skypjack:
That's true but didn't it require the user to create and maintain an object in his/her own space?
If I remember right there was a call like `chain.track_for_unregistration<type>()` to an object that you had to store somewhere. It doesn't differ from creating an `std::vector` as a data member and fill it with function calls like `vec.push_back([]() { entt::meta<T>().reset(); })` for every `T` at the end of the day. We are only replacing a list of calls to `reset` during the shutdown with a list of calls to `track` (or whatever was the name) during the setup.
Am I wrong?

Innokentiy-Alaytsev:
> If I remember right there was a call like chain.track_for_unregistration<type>() to an object that you had to store somewhere.

[There was no such thing](https://github.com/Innokentiy-Alaytsev/entt/blob/chain-based-scoped-meta-reset/test/entt/meta/meta.cpp#L2118)) The user only had to maintain the meta-types registration chain for as long as those types are required. And adding types to the chain was supposed to be done with a call to `entt::meta` with the chain as an argument.

Innokentiy-Alaytsev:
As for creating and maintaining the object - there is always a price. In this case, it doesn't seem too high, especially because the meta-chains are movable and can be concatenated. I don't know why would anyone want it but it's possible and the code for it is very straight forward.

skypjack:
Isn't the user in charge of keeping this alive?
```
auto chain = entt::meta_chain{};
```
This is the equivalent in user space (out of my mind and not tested, just to give an idea):
```
class meta_object_I_must_keep_alive {
    template<typename Type>
    auto init_and_store() {
        auto factory = entt::meta<Type>();
        vec.push_back(+[]() { entt::meta<Type>().reset() });
        return factory;
    }

public:
    void init() {
        init_and_store<a_type>().type("a_type"_hs).data<&a_type::member>();
        init_and_store<another_type>().type("another_type"_hs).data<&another_type::member>();
        // ...
    }

    void deinit() {
        for(auto &&func: vec) {
            func();
        }

        vec.clear();
    }

private:
    std::vector<void(*)()>> vec;
};
```
I've to maintain the object and this is annoying because meta works out-of-the-box with no constraints right now. Moreover, I moved the list of calls to `reset` from the destruction to the construction phase but they are still there.
In a sense, this solution is even better because doesn't rely on static stuff and therefore has no problems with boundaries or so. Didn't the first draft fail for unknown reasons? There were static fields in the chain, it may be the reason behind those errors.

Innokentiy-Alaytsev:
As you can see, there are no problems with my initial proposal. Yes, the user has to store the object of the chain, but the overall design is allocation-free. There are other ways of achieving the same result, but this version seems to be the least restricting.

skypjack:
Ok, test pass, I remembered the opposite and I was wrong but this is a glorified, more verbose and static based version of what I posted in the previous comment.
How does it differ put aside the fact that it's a single linked list rather than a vector?

Innokentiy-Alaytsev:
This way it's provided by the library and allows the user to use something that is already there without adding any restrictions: if the user wants to manage types registration by himself - he can do it. If you don't like the PR - I'm fine with that.

Innokentiy-Alaytsev:
I found out that there are no tests for the cross-boundary case. I'll add them on Monday in the evening. You remember correctly that tests failed for the very first version of PR - it was due to meta-types being registered on first access and the test was checking if the type is available by resolving it with the `entt::resolve<T>()` function. I've removed this because it can never work.

skypjack:
No, well, sorry, I don't want to seem rude. I'm really tired today and I'm having a flight for a business trip that will keep me busy for the whole week. So, fatigue mixed with stress for the next week. Sorry if I look grumpy. 

---

To sum up the reasons for which I'm pushing for a more intrusive approach that doesn't introduce an object to keep alive somehow:

* I wouldn't use this solution because I've not a _meta manager_ around and this is how the whole meta system is designed. The basic idea is that you can use it and forget about it, you don't have to keep alive an object with a context. This would go a bit against this aspect.

* I find it easy to develop user side, especially if it forces me to keep around an object anyway.
  In this case I (as an user) wouldn't have any interest in something built-in that adds more and more static variables if it boils down to an object I must keep alive anyway. I can just store pointers to functions in a vector or something like that. In all cases I can adapt it to my design and do better than having one more indirection over a single linked list.
  To be honest, with a bit of template machinery I think we can reduce it to a single pointer to function to store aside, so why should I use a vector or a static based linked list for that?
  However, this isn't even the most important part. What is important is that usually in `EnTT` I try to avoid providing something that is easy to develop (less then 50/100 loc) user side, that's all. Especially if it's something I won't use. :)

* It's error-prone. `meta<T>()` and `meta<T>(chain)` are both valid calls and you can easily forget to pass the chain object. This means headaches because it seems also pretty hard to spot.

I'm not against providing something to make easier unregistering types but I'd prefer an automatism that fits with the rest of the system and isn't as error-prone as it's today (have you ever forgot to unregister a type? I bet so).

Innokentiy-Alaytsev:
I came up with another idea: it should be possible to tag meta-type nodes
and provide a function for unregistering only the tagged ones. This way
there us no object to store (except for a tag value that can be a constant)
and there us no additional static variables or linked lists. The only
inconvenience is that the user is still required to provide a tag during
registration of each type. It could be solved with some king of RAII and
global state but I don't quite like the latter.

пн, 11 нояб. 2019 г., 3:29 Michele Caini notifications@github.com:

skypjack:
In the meantime I've found a flaw in the chain model. It's due to the static stuff.
I've tested it locally. Let me try to sum up:

* You create a chain and reflect a type in a library:
  ```
  entt::meta<T>(chain_1).type("foo"_hs);
  ```

* Then you initialize another library and extend `T` for _<purposes>_:
  ```
  entt::meta<T>(chain_2).func<&local_func>("func"_hs);
  ```
  This is allowed and it introduces also another problem I'll discuss later.

* Finally, you _destroy_ `chain_2`. The result is that the meta type for `T` is no longer visible from the first library on a system like linux where default visibility isn't hidden.

Of course, we can easily solve it: use `-fvisibility=hidden` or just don't pass the chain during the second call but since it's possible and there are no mechanisms to prevent it, its error-prone nature can be a source of headaches.

This makes evident also another potential problem. Types aren't defined **only** on one side of a boundary. The library allows to _extend_ them on both sides and since it's permitted, someone will do it sooner or later.
Imagine this:

* I define a meta type for `T` within `lib_1`.
* I extend the meta type from `lib_2` by attaching a fake function or a property to it.

Who (and how) should unregister the meta type?
In case `lib_1` unregisters it before `lib_2` goes away, the latter can no longer access the registered function. However, even if the type weren't visible, this is still possible and `lib_2` could expect it to work:
  ```
  entt::meta<T>().func("foo"_hs).invoke();
  ```
In case `lib_2` unregisters the meta type, we have the exact same problem but on the other side of the boundary.
Finally, in case we unload `lib_2` and decide not to unregister the meta type to avoid problems, `lib_1` remains with an apparently valid meta type to which a function is attached that will segfault as soon as it's invoked.

It doesn't sound that good. :)

skypjack:
>it should be possible to tag meta-type nodes and provide a function for unregistering only the tagged ones

Well, this is already possible using properties actually. All we need is a `reset` function exposed also by `meta_type`. Then something like this will just work as expected:
```
entt::resolve([](auto type) {
    if(const auto prop = type.prop("context"_hs); prop && prop.value().cast<ENTT_ID_TYPE>() == "lib_1"_hs) {
        type.reset();
    }
});
```
I must admit this is elegant and can be considered definitely a built-in solution, even though the user has to add the following each and every time:
```
.prop("context"_hs, "lib_1"_hs)
```
Unfortunately, this still suffer from the same problem I mentioned above but I don't see a proper solution to the problem unless one tracks all the meta objects created, not only the meta types.

skypjack:
Btw the more we discuss this, the more I think the problem is the idea of deleting only locally created stuff. The way this library works makes it just impossible. The user is the only one that knows exactly what to do and how to do it.
Either we limit the possibilities (eg you cannot extend a type in a second time but this means you cannot even attach/remove properties, pretty annoying) or we accept that we cannot track locally created stuff easily.

At this point, probably the best thing is really to provide a function to delete everything and rely on the already existing per-type `reset` for all the other cases.

Innokentiy-Alaytsev:
I agree that a general-purpose library like EnTT cannot include solution
for each and every use case. But providing tools for solving typical
problems is a good thing. What about the `reset` function on `meta_type`? I
liked the idea you had about using properties for tagging meta-types. The
snippet you provided is short and simple. It is possible to attach
meta-function for doing the same, but this may result in too much
boilerplate. That said, the `reset` function should not be too hard to
implement. What do you think?

пн, 11 нояб. 2019 г., 12:34 Michele Caini notifications@github.com:

Innokentiy-Alaytsev:
As for "deleting locally created stuff is the problem", for this reason my
initial approach was so explicit - it's the user who is responsible for
knowing what is going on, not EnTT. The tag-based approach has the same
property - it's the user who decides when and on what condition to shoot
one's foot.

пн, 11 нояб. 2019 г., 13:04 Иннокентий Алайцев alaitsev@gmail.com:

skypjack:
>What about the `reset` function on `meta_type`?

We could probably move the `reset` function from the factory to the `meta_type`. This would allow also to search and reset types by name. Moreover, the _reset-all_ function becomes trivial and doesn't require additional functions:
```
entt::resolve([](auto type) { type.reset(); });
```
However, we must solve _a problem_ before to proceed. Currently, the context contains only types with an identifier. All the others aren't reachable. This defeats completely the purpose.

skypjack:
No idea why this have been closed automatically to be honest.
Apparently GH closes issues/PR referred to by commits if you delete the branch that contains those commits.

What about opening an issue to go deeper into the changes I've in mind to match what we discussed above?

Innokentiy-Alaytsev:
Yeah, let's create a new issue and just reference this issue for the contents of the previous episodes.

skypjack:
I mean, unless you can reopen this. GH doesn't allow me to do that. We can continue here but I don't know how to bring this to life.

Innokentiy-Alaytsev:
I don't think I can reopen this PR either. An issue is the best way to go, I think.

skypjack:
Ok, see #361

PR #359, opened 2019-11-16
Added CI for clang-cl on MS VS2019

Innokentiy-Alaytsev:


Innokentiy-Alaytsev:
I've changed the base to `master` because build on `wip` fails due to some tests failure.

Innokentiy-Alaytsev:
Current problems:
1. I don't know how to perform multiple different builds per platform, i.e. how to build both with MSVC and Clang-CL.
2. It does use Clang-CL for building but fails to build anything because of build-time errors. It's kinda hard to read through all the warnings (due to `-Wall` and `-pedantic`). From what I managed to understand (after 5 minutes of FF freeze) is that for some reasons tests are built without exceptions support.
3. My local build partially succeeds. The only test that fails to build is the one for `emitter` - the compiler is not happy with `noexcept` specifier on defaulted constructor of the `entt::emitter`.

skypjack:
>My local build partially succeeds. The only test that fails to build is the one for emitter - the compiler is not happy with noexcept specifier on defaulted constructor of the entt::emitter.

Because that's wrong! Good point. Default constructor for `std::vector` isn't `noexcept` actually. Thank for pointing this out.
I'll try to merge your PR with the current workflow to make it work with GH Actions. :+1:

skypjack:
See branch `wip`.

Issue #368, opened 2019-12-03
Serialisation Question


ivanhawkes:
First up, I love the work you've done. I'm working on implementing it in my RPG project built using CRYENGINE. That leads to me having a few more constraints than some projects, which have led to me asking this question.

I am trying to serialise the entities and components out to disk. Sadly, XML is the only format I've got a decent chance of using due to the state of the libraries I am working with. Even then, it's a mess. The code to write an XML file is done, so I only need to read it in now, and that's 80% done.

I have some test output XML:

```
<entities>
 <entity>
  <components>
   <position typeGUID="d22ce389-15df-4ef3-b6cf-fd7490900765">
    <properties CryXmlVersion="2" X="20.799999" Y="20.799999"/>
   </position>
   <velocity typeGUID="812ddc29-d565-4294-8a1d-ce5af770151b">
    <properties CryXmlVersion="2" DX="0.80000001" DY="0.80000001"/>
   </velocity>
  </components>
 </entity>
 <entity>
  <components>
   <position typeGUID="d22ce389-15df-4ef3-b6cf-fd7490900765">
    <properties CryXmlVersion="2" X="15.6" Y="15.6"/>
   </position>
   <velocity typeGUID="812ddc29-d565-4294-8a1d-ce5af770151b">
    <properties CryXmlVersion="2" DX="0.60000002" DY="0.60000002"/>
   </velocity>
  </components>
 </entity>
</entities>
```
The idea is simple enough, read in the entities, make a new entity with EnTT, read each component - and use the GUID to figure out which class type I need to instantiate. That's where I am stuck.

Is there already a method to register types somewhere with a key and then have one created by passing in the key? Is this even possible? I don't mind having to register the components with something, as they already have to do that to work with CRYENGINE.

Is there another way to handle this? I did look at the snapshot code, and ran into a few issues as I tried to test it. My archive types didn't seem to have the right functions or something along those lines. I probably need to put more time into that.

I tried using Cereal as a quick and dirty way to get started with the testing but it doesn't compile within my project with the restrictions of CRYENGINE...which for some reason has decided to #define malloc and a few other crimes against good code. It relies on RapidJSON and RapidXML too, which are both looking a little abandoned - plus, I would be adding a third serialisation library to the code >.<

Ideally, I'd just be able to use a factory to create the components, one that is key based (templated). I'm just not sure if you already support that, or how it would work if not supported.

Innokentiy-Alaytsev:
Hello!

[...]

PR #373, opened 2019-12-15
Fix MSVC non-member template functions compilation error

suVrik:
Hey! With a recent MSVC update my project fails to compile because I use non-member template functions in EnTT reflection. For whatever reason MSVC considers these functions to be references rather than pointers. EnTT tests fail to compile as well, even though non-member template `func` is not covered (non-member template setters and getters are). Here's the fix for this issue. I added tests for both member template functions and non-member ones just in case of future MSVC surprises. Cheers

### DIFF
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -90,7 +90,7 @@ bool setter([[maybe_unused]] meta_handle handle, [[maybe_unused]] meta_any index
     bool accepted = false;
 
     if constexpr(!Const) {
-        if constexpr(std::is_function_v<std::remove_pointer_t<decltype(Data)>> || std::is_member_function_pointer_v<decltype(Data)>) {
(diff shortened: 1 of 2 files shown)


skypjack:
Good catch, thank you for pointing this out. May I ask you what version of MSVC is it?
Feel free also to add you nick to the list of contributors if you want (file AUTHORS, lexicographic order).

suVrik:
I'm already on the list! My MSVC version is 2019 16.4.1. Compiler version is 14.24.28314.

skypjack:
Oh, sorry, I'm from mobile and I didn't check before writing the message. I apologize.

skypjack:
Since the problem is on the current `master` and also on the last tag (see #368), I've decided to cut a new patch release of `EnTT`.
However, this PR has some conflicts with the work done on `experimental`, that is the upcoming `master` (hopefully by the end of the month).

My plan is to port only the relevant changes to the `.hpp` files on `master` so as to reduce the conflicts, then add the tests directly on the rework from `experimental`.
Put simply, the new tests strengthen the test suite but they aren't strictly necessary for the patch since it's easily reproducible with the current tests. So we can safely split the whole thing in two parts. :+1:

PR #391, opened 2020-01-13
entity/helper: add invoke_on_component helper

### LINKED ISSUE #385: Member function for on_construct() without instance?
Would it be possible to add an overload for the `on_construct()`-family of methods on the sinks to allow for a member function of the underlying component type, without an instance of the member function's type?

Here's a visual example:

```c++
struct some_component {
    int i = 0;

    void init() {
        cout << "I've init'd: " << i << endl;
    }
};

registry.on_construct<some_component>().connect<&some_component::init>();
```

In the above pseudo-code, we don't use a static function nor do we use a member function with a point/reference to a specific instance. Instead, we use a pointer to a member function _of the component itself_ to call, with `this` being the pointer to the newly created component itself.

This would greatly clean up some of the glue code in some parts of the engine, as right now I am either passing lambdas or creating stubs to do pretty much exactly this.

For context, I could be using constructors and destructors, but given the inherently POD-like structures, default move constructors, et al, I'm writing my components in a way they can be re-used via `init()` and `destroy()` member methods (where a call to `destroy()` could, in theory, be followed by `init()` assuming all of the non-default POD properties have been set prior). I fully realize this isn't a design requirement by ENTT but it seems like a clean approach for my use-case.

Would love to hear your thoughts, or why this might not be a good idea. :)

Qix-:
Closes #385.

The name `invoke_on_component` is still [up for debate](https://github.com/skypjack/entt/issues/385#issuecomment-573482861) - as is this entire PR. Feel free to close, was just something I felt would be helpful. :)

### DIFF
--- a/src/entt/entity/helper.hpp
+++ b/src/entt/entity/helper.hpp
@@ -133,6 +133,33 @@ template<ENTT_ID_TYPE Value>
 using tag = std::integral_constant<ENTT_ID_TYPE, Value>;
 
 
+namespace internal {
+
+template <typename Arg, typename... Args>
+auto component_type(void (*)(Arg, Args...)) -> std::decay_t<Arg>;
+
+template <typename Class, typename... Args>
+auto component_type(void (Class::*)(Args...)) -> Class;
+
+template <typename Class, typename... Args>
+auto component_type(void (Class::*)(Args...) const) -> Class;
+
+}
+
+template <auto Func, typename Entity = entt::entity>
+void invoke_on_component(const Entity entity, entt::basic_registry<Entity> &reg) {
+    using Component = decltype(internal::component_type(Func));
+    // decltype(auto) instead of Component & to handle the empty component optimization
+    decltype(auto) comp = reg.template get<Component>(entity);
+    if constexpr (std::is_invocable_v<decltype(Func), decltype(comp), Entity, decltype(reg)>) {
+        std::invoke(Func, comp, entity, reg);
+    } else if constexpr (std::is_invocable_v<decltype(Func), decltype(comp), Entity>) {
+        std::invoke(Func, comp, entity);
+    } else {
+        std::invoke(Func, comp);
+    }
+}
+
 }
 
 
--- a/test/entt/entity/helper.cpp
+++ b/test/entt/entity/helper.cpp
@@ -47,3 +47,54 @@ TEST(Helper, Tag) {
 
     ASSERT_EQ(counter, 0);
 }
+
+struct my_component {
+    mutable int *foo;
+
+    static void create_s(my_component &c)                   { *(c.foo) |=  0b00000001; }
+    void create()                                           { *(  foo) |=  0b00000010; }
+    void create_e(entt::entity)                             { *(  foo) |=  0b00000100; }
+    void create_er(entt::entity, entt::registry &)          { *(  foo) |=  0b00001000; }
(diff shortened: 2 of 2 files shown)


Qix-:
@Kerndog73 on GCC/Linux:

```
/home/runner/work/entt/entt/src/entt/entity/helper.hpp: In function ‘void entt::invoke_on_component(Entity, entt::basic_registry<Entity>&)’:
/home/runner/work/entt/entt/src/entt/entity/helper.hpp:153:44: error: expected primary-expression before ‘>’ token
     decltype(auto) comp = reg.get<Component>(entity);
```

Not sure what it's complaining about.

Also, did I cover every case?

Kerndog73:
> Not sure what it's complaining about.

`get` is a dependent name. I'm pretty sure clang should have thrown an error. Try this:

```C++
decltype(auto) comp = reg.template get<Component>(entity);
```

> Also, did I cover every case?

You forgot the free function case.

Kerndog73:
BTW, I like how you used a bitset. Might I suggest using binary literals (`0b100`) to make it a little clearer.

Qix-:
No idea how I didn't know about binary literals. You're blowing my mind today @Kerndog73 :)

Updated, let me know how that looks.

Kerndog73:
Maybe it wants you explicitly take the address of the function `&entt::invoke_on_component`? I think gcc is wrong in that case. Functions should decay to function pointers. Or maybe that's not the case for template parameters. I'm no language lawyer.

Kerndog73:
I messed around a bit on compiler explorer and I think I figured it out. It didn't like that you were declaring a static member function inside of a function. Try moving `my_component` into global scope.

Qix-:
Good solve, that seems to have fixed it :)

Kerndog73:
Hooray! 🎉

Qix-:
> We found two compiler bugs in one little PR. I can't imagine the pain Michele has to go through!

Hehe, newer standards tend to find weird fringe cases like that. Entt looks like some of the more advanced C++ I've seen utilized, so it doesn't surprise me. It's good that all of the cross-platform CI is set up!

skypjack:
On hold because of [this](https://github.com/skypjack/entt/issues/385#issuecomment-573872132) comment.

skypjack:
I looked into the code. Unfortunately, this doesn't work with all signals.
In particular, `on_replace` has a different signature and we couldn't attach `on_component` to it.

Kerndog73:
Would it even make sense in that case? `invoke_on_component` should return one component so which component would it get from `on_replace`? Maybe there could be a separate version for the `on_replace` signal or just leave `on_replace`.

skypjack:
I don't know if it even makes sense actually. :smile: 
The fact is that a _get-and-forward_ approach like this one cannot work with the `on_replace` because the new component is passed as an argument, so 1) I change this behavior but we have the opposite problem and we can't get the _old_ instance now, 2) we write a _dedicate_ function or 3) who cares.

Qix-:
Hmm. Yeah. That's a problem.

FWIW I didn't face this because I don't use component replacement, never saw the need. I usually attach components to an entity and they stay there for the course of the entity's lifetime.

skypjack:
I'm closing this because it doesn't cover all the possible requirements.
If it's incomplete, it doesn't worth it. This is more or less the policy of this library. :wink: 
I'll try to implement something soon. I'll give you more updates in the issue you opened on the topic. :+1:

Issue #407, opened 2020-02-05
empty components causing warning in registry::replace


RafeWoo:
First of all, thanks for your great library. 
I'm currently playing around with it and enjoying myself so far.

I have hit a problem though. I just switched from #including the single header to individual headers. 
and the code stopped compiling due to this warning

    entt\entity\registry.hpp(123): warning C4172: returning address of local variable or temporary

The warning only occurs if I try to assign empty components like a tag to an entity.

The first thing I noticed was that the code in the basic_registry::pool_handler::replace function is different between the single_include and individual registry.hpp. In the single header there is a test for empty types. (Are there any plans to keep the single header in sync?)
It looks like in December 2019 the ETO check was removed quite intentionally but if I add it back in the code seems to work.
  
    template<typename... Args>
        decltype(auto) replace(basic_registry &owner, const Entity entt, Args &&... args) 
    {
    #ifdef USE_ETO_TEST
            if constexpr (ENTT_ENABLE_ETO(Component)) {
                ENTT_ASSERT((storage<Entity, Component>::has(entt)));
                update.publish( owner, entt, Component{});
                return Component{ std::forward<Args>(args)... };
            }
            else {
                Component component{ std::forward<Args>(args)... };
                update.publish(owner, entt, component);
                return (this->get(entt) = std::move(component));
            }
    #else
             Component component{std::forward<Args>(args)...};
             update.publish(owner, entt, component);
             return (this->get(entt) = std::move(component));
    #endif
    }

I'm using the latest master build.
I'm using VS2017 and compiling with warnings as errors.

What I think is happening is this:  On the last line, this->get() is returning by value for empty structs,  then the assignment operator is returning a reference to the local value which is being returned from the replace function as a whole triggering the warning.

Thank you again.

skypjack:
Afaik it's an issue of VS2017 that has been solved in VS2019.
Actually, the function doesn't return a reference to a local variable. The warning is wrong.
I'd suggest either to update your VS or not to treat warnings as errors. However, if you know how to suppress it, please let me know and I can apply your changes to `master`.
Since I'm not working on Windows and even less with VS2017, I've some problems to experiment in this sense.

RafeWoo:
Thanks for prompt response. I've added a pull request that fixes the warning by adding a replace method to the storage class.

[...]

PR #408, opened 2020-02-06
Fix for #407 - vs2017 warning:  warning C4172: returning address of local va…

RafeWoo:
Fix for #407 :  vs2017 warning:  warning C4172: returning address of local variable or temporary

When given an empty type registry::pool_manager::replace calls storage::get which returns by value,
operator= then returns a reference to that temporary value which is then returned from replace triggering the warning.
I moved the assignment code from registry::pool_manager::replace down into storage class by adding a replace function to storage
This means the 'assignment' is done before returning the value or ref from the storage class.

### DIFF
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -1,4 +1,4 @@
-﻿#ifndef ENTT_ENTITY_REGISTRY_HPP
+#ifndef ENTT_ENTITY_REGISTRY_HPP
 #define ENTT_ENTITY_REGISTRY_HPP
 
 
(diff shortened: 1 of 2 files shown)


skypjack:
This PR fails on all systems. Moreover, it's probably bigger than what is actually needed to solve #407.
I've pushed a small change on `experimental`. The CI no longer shows the warning for VS2016.
May I ask you to make your tests and confirm it's solved eventually? Thank you very much.

Please, don't close neither this PR nor the issue associated with it.
Both will be closed when the commit is merged on `master`. :+1:

Innokentiy-Alaytsev:
Will putting something like following just before returning the value work?
```c++
#if defined(_MSC_VER)
#   define ENTT_LOCAL_VARIABLE_ADDRESS_RETURN_SUPPRESS_WARNING                                \
        __pragma (warning (suppress : 4172))
#else
#   define ENTT_LOCAL_VARIABLE_ADDRESS_RETURN_SUPPRESS_WARNING
#endif
```

skypjack:
@Innokentiy-Alaytsev if you look at the commit, it's shorter than this. :smile:

PR #418, opened 2020-02-17
Create CODE_OF_CONDUCT.md

skypjack:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
```
OvermindDL1: This doesn't seem fully inclusive.  Why not instead say something like `harassment-free experience for everyone.` (full stop) as then that is fully inclusive instead of some arbitrary traits that have nothing to do with ones ability to program?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
```
OvermindDL1: Using examples implies these are the hard lines to follow, meaning people can find ways to work around it.  Why not instead just say something along the lines "Any behaviour that does not help foster an open and welcoming community that focuses on the code of the project is unacceptable.", or something along those lines.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
```
OvermindDL1: This is way *way* too generic.  It is possible for anyone to call anything 'harassment', even if it is as simple as rejecting a PR due to it not fitting in the project.  This is exceptionally open to abuse even more so than the rest of this file seems to be.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
```
OvermindDL1: Relating to a professional setting, a lot of the document is not very professionally worded.

OvermindDL1:
This is a very odd thing to see appear here.  I'm unsure what value it brings at the best (have trolls been an issue here that are not already easily dealt with GitHub's tools, and how would this stop them anyway?), and it seems extremely open to abuse at worst, which is something that does not belong in a programming project.

skypjack:
Hold on @OvermindDL1 I was trying to setup issue templates (because I often ask the same details over and over and these can help users to provide the right information from comment 0) and saw this _code of conduct_. I clicked to see what was the result but I put it in a branch and didn't really want to merge it since it's _too much_ for such a small project imho.
Thanks for your feedback though. They pointed out some oddities I wouldn't have noticed otherwise. :+1:

OvermindDL1:
I actually did further research on this specific CoC after my post and wow but that one is reviled by a *lot* of the FOSS community due to it's abuse, to put it mildly...  ^.^;

There are better CoC's out there I found if you are actually curious though, but honestly I've never seen the point.  Trolls will be trolls, just block and go, everyone else that is worth teaching should be taught is my methodology, never ignore any question, everything can be a learning opportunity.  :-)

PR #428, opened 2020-02-25
replace all misspelt to_integral() with to_integer() but remain to_integral() marked deprecated

qiujiangkun:
replace all misspelt to_integral() with to_integer() but remain to_integral() marked deprecated

### DIFF
--- a/single_include/entt/entt.hpp
+++ b/single_include/entt/entt.hpp
@@ -1462,7 +1462,7 @@ using tag = std::integral_constant<ENTT_ID_TYPE, Value>;
  */
 #define ENTT_OPAQUE_TYPE(clazz, type)\
     enum class clazz: type {};\
-    constexpr auto to_integral(const clazz id) ENTT_NOEXCEPT {\
+    constexpr auto to_integer(const clazz id) ENTT_NOEXCEPT {\
         return static_cast<std::underlying_type_t<clazz>>(id);\
     }\
     static_assert(true)
@@ -2819,7 +2819,7 @@ using tag = std::integral_constant<ENTT_ID_TYPE, Value>;
  */
 #define ENTT_OPAQUE_TYPE(clazz, type)\
     enum class clazz: type {};\
-    constexpr auto to_integral(const clazz id) ENTT_NOEXCEPT {\
+    constexpr auto to_integer(const clazz id) ENTT_NOEXCEPT {\
         return static_cast<std::underlying_type_t<clazz>>(id);\
     }\
     static_assert(true)
@@ -3943,7 +3943,7 @@ public:
 
     template<typename Entity>
     constexpr bool operator==(const Entity entity) const ENTT_NOEXCEPT {
-        return (to_integral(entity) & traits_type<Entity>::entity_mask) == to_integral(static_cast<Entity>(*this));
+        return (to_integer(entity) & traits_type<Entity>::entity_mask) == to_integer(static_cast<Entity>(*this));
     }
 
     template<typename Entity>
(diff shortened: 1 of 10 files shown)


skypjack:
It's `to_integral` as in `integral_type`. `int` as in `integer` is just one of the possible integral types along with long, long long, short, char, bool and so on. I think `to_integral` is a better fit that `to_integer`. What's the reason for changing the name exactly?

OvermindDL1:
Just for note:

> in·te·gral
> /ˈin(t)əɡrəl,inˈteɡrəl/
> adjective
>
> MATHEMATICS
> of or denoted by an integer.

So `integral` is correct.  It's not an integer, but it's denoting an integer.

qiujiangkun:
I didn't notice integral_type, but I think integral_type is OK. As an adjective, integral refers to both integers and integration, which is a bit confusing. However, in to_integral(), integral is obviously a noun, meaning the result of an integration

```text
in·te·gral
I. \ˈin-ti-grəl (usually so in mathematics); in-ˈte-grəl also -ˈtē- also ÷ˈin-trə-gəl\ adjective
 Date: 1551
 1. 
   a. : essential to completeness : constituent 
     <an integral part of the curriculum>
   b. 
     (1) : being, containing, or relating to one or more mathematical integers
     (2) : relating to or concerned with mathematical integrals or integration
   c. : formed as a unit with another part 
     <a seat with integral headrest>
 2. : composed of integral parts
 3. : lacking nothing essential : entire
 • in·te·gral·i·ty \ˌin-tə-ˈgra-lə-tē\ noun
 • in·te·gral·ly \ˈin-ti-grə-lē; in-ˈte-grə- also -ˈtē-\ adverb
II. noun
 Date: circa 1741
  : the result of a mathematical integration — compare definite integral, indefinite integral
```

qiujiangkun:
Moreover, there's a `std::to_integer()`, so changing to_integral() to to_integer() makes more sense of continuity

skypjack:
I'm not a native speaker. You two are confusing me. :smile: 
I'll leave this kind of discussion to you and @OvermindDL1 I can only learn something this time, I don't think I can add much instead.

skypjack:
Btw it's not `to_integer`, it's `to_integer<T>` where you specify the desired type.
In this case, it's more similar to an hypothetical `to_underlying` rather than `to_integer<T>` since the type is fixed.

qiujiangkun:
Neither am I. I just pay some more attention to these trivial names. I was looking for a method to inspect the inner structure of an entity, and someone let me use `to_integer()`.  However,  what I only found was to_integral(), instead of to_integer(). I thought it was a misspelling in the first place.

qiujiangkun:
> Btw it's not `to_integer`, it's `to_integer<T>` where you specify the desired type.
> In this case, it's more similar to an hypothetical `to_underlying` rather than `to_integer<T>` since the type is fixed.

Really? What I found was this function without templates. Users should use a macro `ENTT_OPAQUE_TYPE(clazz, type)` to specify in the inner type of an entity.
```cpp
constexpr auto to_integral(const clazz id) ENTT_NOEXCEPT
{
...
}
```

qiujiangkun:
I googled more just to find that many languages and libraries use to_integer() while far more languages and libraries use to_integral() based on the numbers of results. 46,300(to_integer()) vs  2,040,000(to_integral()).
Should have I closed this PR.

OvermindDL1:
Honestly I'd just leave it as it is, I see `integral` used in similar contexts elsewhere often enough that it seems clear either way, and not changing it means not surprising users of this library (and I'm not a fan of keeping deprecated functions, even just trampolines, as it adds in more vectors for bugs).  ^.^

skypjack:
Thanks for reporting and for discussing it @qiujiangkun really appreciated. :heart:

PR #430, opened 2020-02-26
Re-added bazel support

zaucy:
Hey it's me again the bazel guy.

I was updating my project to the latest version of EnTT (v3.3.0) and noticed it no longer worked with bazel! I see it was removed due to it not being maintained. I'll admit I hadn't been running the tests so I didn't notice that shortly after it was initially added the tests no longer worked with bazel. However my project had been running EnTT v3.2.2 just fine because the user facing target `@com_github_skypjack_entt//:entt` had been working.

So that brings us to this PR. This PR includes only the user facing target and no tests to make it simpler.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,15 @@
+_msvc_copts = ["/std:c++17"]	
+_gcc_copts = ["-std=c++17"]
+
+cc_library(
+    name = "entt",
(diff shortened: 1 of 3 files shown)


skypjack:
That's great. Since I'm not using bazel, I didn't know how to update it and things got broken quickly. Thank for making it work again. :+1:

PR #433, opened 2020-03-01
Rework assign in registry

FabienPean:
Hello, I reworked the assign method in the registry to allow more options. I vaguely remember a comment on gitter about something like that. Method is backward compatible, using C++17 and should work on all main compilers.

It enables the following calls
```c++
auto&& c = registry.assign<Component>(entity, Args...);
auto&& c = registry.assign(entity, Component{Args...});
auto&&[c1,c2] = registry.assign(entity, Component1{Args1...}, Component2{Args2...})
```
Feel free to rework the branch as needed.

### DIFF
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -620,10 +620,19 @@ public:
      * @param args Parameters to use to initialize the component.
      * @return A reference to the newly created component.
      */
-    template<typename Component, typename... Args>
+    template<typename Component = void, typename... Args>
     decltype(auto) assign(const entity_type entity, Args &&... args) {
         ENTT_ASSERT(valid(entity));
-        return assure<Component>().assign(*this, entity, std::forward<Args>(args)...);
+		if constexpr (std::is_same_v<Component, void> && sizeof...(Args) > 1) {
+			static_assert((!std::is_lvalue_reference_v<Args> && ...), "registry.assign(...) does not accept lvalue reference");
+			return std::tuple<std::remove_cv_t<std::remove_reference_t<Args>>&...>{ assure<Args>().assign(*this, entity, std::forward<Args>(args))... };
+		} else if constexpr (std::is_same_v<Component, void> && sizeof...(Args) == 1) {
+			static_assert((!std::is_lvalue_reference_v<Args> && ...), "registry.assign(...) does not accept lvalue reference");
+			return (assure<Args>().assign(*this, entity, std::forward<Args>(args)),...);
+		}
+		else {
+			return assure<Component>().assign(*this, entity, std::forward<Args>(args)...);
+		}
     }
 
     /**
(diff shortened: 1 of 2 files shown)


skypjack:
Uhm... I think `registry.assign(entity, Component1{}, Component2{})` may conflict with the already existent component-less overload for `assign`. It's not that easy to make it conflict actually but we are probably overloading this too much. :smile: 

Is the goal to create an overload of assign that doesn't require to specify the component type?
Because `registry.assign<T>(entity, T{args...})` works already afaik.

FabienPean:
I must admit I didn't check other overload and relied on test to catch it if something was breaking. 😄 

Yes, the goal is to drop specifying the component type.

skypjack:
I'm not sure we can merge this as-is actually. I'll give it a try soon though. :+1:

Kerndog73:
I was looking at the **FOUR** overloads of `assign` and I couldn't wrap my head around the difference between [642](https://github.com/skypjack/entt/blob/50ba8c6c396f69e31876e7e4a248f3a0a7b2a819/src/entt/entity/registry.hpp#L642) and [661](https://github.com/skypjack/entt/blob/50ba8c6c396f69e31876e7e4a248f3a0a7b2a819/src/entt/entity/registry.hpp#L661). The function bodies are the same but somehow they do different things. Perhaps some of overloads could be given different names? [678](https://github.com/skypjack/entt/blob/50ba8c6c396f69e31876e7e4a248f3a0a7b2a819/src/entt/entity/registry.hpp#L678) has nothing to do with components yet it has the same name as three functions that do. But they're still assigning something so... I don't know. It's just a suggestion.

---

As for the PR, well I don't think it's necessary. `assign` uses aggregate initialization (unlike `emplace_back` from the standard library). So your second could be rewritten using the existing function.

    auto&& c = registry.assign(entity, Component{a, b, c});
    auto&& c = registry.assign<Component>(entity, a, b, c);

For the third example, it just looks like a really long line. I can see how it might make sense for assigning multiple tags in one line (I think I actually requested that long ago!). However, my personal preference has shifted since then and I think it's cleaner to put things on separate lines.

FabienPean:
Indeed, it is not *necessary*, just an alternative assign that is convenient. The goal was to get rid of the explicit template parameter while avoiding breaking things.

I managed to find the message that made the inception 😄 
> Michele Caini @skypjack Jan 12 23:18
> btw I think the initial design of EnTT is wrong on an aspect
> this one: `registry.assign<T>(entity, 2, 'c', 42.)`
> should have been : `registry.assign(entt,T{2, 'c', 42.});`

Feel free to do whatever with the PR, that was a fun exercise!

skypjack:
Ok, so, these two functions are now available on `master` as `emplace` and `insert`:
```
auto&& c = registry.assign<Component>(entity, Args...);
auto&& c = registry.assign(entity, Component{Args...});
```
I prefer not to include this one instead:
```
auto&&[c1,c2] = registry.assign(entity, Component1{Args1...}, Component2{Args2...})
```
I'd rather prefer to invite users to use `emplace` or `insert` for ranges since they can be faster than passing N instances of different components.

I'm closing this PR since there is nothing to do on `master`.
I didn't merge the PR directly because many functions are undergoing a renaming to disambiguate some calls.

PR #436, opened 2020-03-04
add edge case test for groups

pgruenbacher:
This test case currently fails on master. seems to fail at here:

https://github.com/skypjack/entt/blob/50ba8c6c396f69e31876e7e4a248f3a0a7b2a819/src/entt/entity/registry.hpp#L1412-L1414


skypjack:
Apparently multi component view's iterators don't like if we start swapping elements during iterations. :smile:

skypjack:
Ok, I've found the issue, the reason and the fix.
I'll push it upstream soon and make a new release.
Thanks for pointing this out!!

PR #444, opened 2020-03-11
Calling a meta function with entt::meta_any as parameter

MatthieuMv:
Hello !

This patch check in the parameter tuple constructor if the target parameter type is **entt::meta_any** thus skip any conversion.

This simple example shows the issue that I tried to fix. If you try to run it, you will see that foo is never called.
```C++
#include <iostream>
#include <entt/meta/meta.hpp>
#include <entt/meta/factory.hpp>

struct MyStruct
{
    void foo(const entt::meta_any &any) { std::cout << "Foo called " << any.cast<int>() << std::endl;}
};

int main(void)
{
    using namespace entt; // '_hs' hash literal

    entt::meta<MyStruct>()
        .alias("MyStruct"_hs)
        .func<&MyStruct::foo>("foo"_hs);

    MyStruct myStruct;
    entt::meta_any var { 42 };

    entt::resolve<MyStruct>().func("foo"_hs).invoke(myStruct, var);
}
```

### DIFF
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -212,6 +212,8 @@ meta_any invoke([[maybe_unused]] meta_any instance, meta_any *args, std::index_s
     [[maybe_unused]] const auto direct = std::make_tuple([](meta_any *any, auto *value) {
         using arg_type = std::remove_reference_t<decltype(*value)>;
 
+        if constexpr (std::is_same_v<arg_type, entt::meta_any>)
+            return any;
         if(!value && any->convert<arg_type>()) {
             value = any->try_cast<arg_type>();
         }


skypjack:
You aren't passing the object in the outer context though.
Instead, the function receives a copy of it. Do you know that?

MatthieuMv:
For what I've learnt from the meta library, when invoking a function, variadic arguments are forwarded to a tempory **any** array.
Then, the factory's **invoke** method creates a tuple containing all well formated arguments before function call.

If I give an **int** or an **any** containing an **int**,  they are both stored in the first array as **any** to **int**.
The problem occurs in the second step where the system tries to convert the type **int** to the type **any** (as required in function signature).
The function is thus never called (like the example code).

This is why in the patch I check at compile time if the target type is **any**, and then skip the conversion.

skypjack:
>when invoking a function, variadic arguments are forwarded to a temporary any array

Not exactly.
You can wrap arguments in a `std::ref` to pass them by reference to the function (so as to edit objects in the outer context) and you can deref a `meta_any` (`*any`) in order to do the same rather than triggering a copy.
In other words, you can **always** decide if you want to pass things by value or by reference.
This is no longer true in your case. The `meta_any` is always copied and it's a shame because it means that the function cannot _modify it_ and then let you use the modified value directly.

skypjack:
Oh, btw `_hs` literal is already in the outer scope, no need for an `using namespace entt;`. :wink:

MatthieuMv:
Thanks to your comment I realised that the following code works perfectly :

```C++

#include <iostream>
#include <entt/meta/meta.hpp>
#include <entt/meta/factory.hpp>

struct MyStruct
{
    void foo(entt::meta_any &any) {
        any = 43;
        std::cout << "Foo called " << any.cast<int>() << std::endl;
    }
};

int main(void)
{
    entt::meta<MyStruct>()
        .alias("MyStruct"_hs)
        .func<&MyStruct::foo>("foo"_hs);

    MyStruct myStruct;
    entt::meta_any var { 42 };

    entt::resolve<MyStruct>().func("foo"_hs).invoke(myStruct, entt::meta_any(std::ref(var)));
    std::cout << "Outter scope var is " << var.cast<int>() << std::endl;
}
``` 
Producing :
```console
Foo called 43
Outter scope var is 43
```

My patch breaks the outer reference in this case and it doesn't change the result so we have :
```console
Foo called 43
Outter scope var is 42
```

I close the PR since it isn't a valid fix.

However we still can't invoke a function taking a copy/ref to **any** with a parameter that isn't a reference_wrapper to an **any** because the conversion system doesn't know how to deal with it.
A function call with a copy is better than no function call at all (even more when dealing with scalar types where copying is not a big deal).
Do you know any way to fix it ?

Anyway, thank you for your help !

skypjack:
I was to reply with what you asked! :smile: 
I must admit that registering a function that requires a `meta_any` as an argument wasn't an intended use initially, since `meta_any` is mainly used to push things back and forth across the opaque boundary of the meta system.

That said, as you pointed out this is the way to go to pass the `meta_any` _by reference_:
```
entt::resolve<MyStruct>().func("foo"_hs).invoke(myStruct, entt::meta_any{std::ref(var)});
```
You can't pass the `std::ref` directly here because of how it works and the rules of the language. Since the reference wrapper **can** convert to a `meta_any`, the implicit initialization isn't viable. Fortunately, the possibility of doing it explicitly gets around the issue.

To pass the `meta_any` _by value_ instead, we have a trickier syntax:
```
entt::resolve<MyStruct>().func("foo"_hs).invoke(myStruct, entt::meta_any{std::in_place_type<entt::meta_any>, var});
```
The _problem_ is that all arguments are used to construct an array of `meta_any` within `invoke. Because of this, if you pass your instance directly, it's copied and it's meta type remains that of the underlying object.
What you want is to create a `meta_any` that contains another `meta_any` instead. To do that you must get around the copy. The only way is to explicitly create this kind of objects before passing anything to the `invoke`. `entt::meta_any{std::in_place_type<entt::meta_any>, var}` does exactly this, then the resulting `meta_any` is copied internally but its underlying type doesn't change, nor the contained object that is passed by value this time though.

So, all in all, you **can** pass `meta_any`s to opaque functions, it's just trickier than with any other type because they are also the mean used to bring values from your context, that is _explicit, to the opaque one.

MatthieuMv:
Thank you for these interesting explanations.

I have now a better understand of my issue and why we need to be explicit about **meta_any** instances.

Let me explain you my use case :
I'm currently building a C++ voxel game framework and I need to write a high-level object-oriented interpreted language like Qt's one ([QML](https://doc.qt.io/qt-5/qmlfirststeps.html)) for front-end development. I don't want to use external compilation tools for the reflection (like Qt's [MOC](https://doc.qt.io/qt-5/why-moc.html)) and your library perfectly fits my need since it's easy to integrate and build on top of it.

With the current version I can now easily create the internal system of the interpreter but as this framework aims to be as intuitive as possible, I'd like to reduce miss-usage cases to the minimum.

As I don't know any way to check if the function call has failed, I can't even tell when the user isn't invoking correctly.

I think a *dirty* workaround would be to create another **any** like type but I'd have to specify a conversion function for every other type registered to be auto convertible.

Anyway thank you for the time invested, I'm definitely enjoying learning with your library and your comments :)

skypjack:
>As I don't know any way to check if the function call has failed, I can't even tell when the user isn't invoking correctly.

What do you mean with _the function call has failed_? You can know it actually with this system.

In fact, `meta_any` has a particular state that we can call _uninitialized_. Meta functions returns uninitialized `meta_any`s when something goes _wrong_.
Note also that `meta_any` supports **all** types, even `void`. Therefore, a function returning `void` generates a correctly initialized `meta_any` in case everything works as expected, similarly to what happens for functions returning types other than `void`. Ofc you won't be able to _get the value_ from a valid `meta_any` that _contains_ a `void` but still it's a valid `meta_any`.

So, you can do this with **all** meta functions to know if they have _failed_, also with those that wrap functions returning a `void`:
```
if(auto any = func.invoke(obj, args...); any) {
    // ...
}
```
The `operator bool()` of a `meta_any` returns `true` if it's correctly initialized.
I assumed that a failing functions is one for which `invoke` fails because for example you passed the wrong argument.

>I'm definitely enjoying learning with your library and your comments :)

I'm glad to know that. I hope the answer above has solved your problem. :+1:

MatthieuMv:
I tried to implement it this way with separate **meta** library but caught errors when returning from void function. 
I probably missed something when doing it..

I think for now I'll forget ambiguous calls to any and just check when invoking, thank you :)

skypjack:
Standalone `meta` has not this feature. That's why I invited you to use the meta module from `EnTT`. :wink: 
You can freely return from `void` functions and the `meta_any` you receive contains enough information to disambiguate all situations. :+1:

Issue #446, opened 2020-03-16
API for removing all components from an entity


DonKult:
Hi,
lets say I have an entity I want to reload without modifying other entities which refer to this one by id. For this I would first like to strip all components from the entity but that already poses a problem:

Knowing what `registry.clear()` vs `registry.clear<Component>()` does I thought `registry.remove(entity)` might do it – turns out it doesn't, it is a no-op instead. `registry.visit(entity, [](auto const component){…})` might have worked if I could have figured out a way of doing the equivalent of `registry.remove<component>(entity)` with a `component` only known at runtime. All this runtime machinery seems rather powerful, but sadly I couldn't find too much documentation on how to make that actually work in practice. What seems to work is `registry.destroy(entity); assert(entity == registry.create(entity));`, but is that the way it should be done? I feel kinda uneasy about it for some reason. I had no qualms if that API were `registry.destroy(entity); assert(entity == registry.recycle(entity))` instead so I know its a baseless fear, but still. I can also not remove the feeling of a previous and by now probably ancient version having a `registry.clear(entity)` for this propose, but I haven't dug into history to checked. 

So this is as much a question of how this should be done nowadays as it is a request to fix and/or at least document the pitfalls a user might encounter along the way. Thanks!

skypjack:
>All this runtime machinery seems rather powerful, but sadly I couldn't find too much documentation on how to make that actually work in practice.

Because it entered `EnTT` with the last release and only in the form of the `visit` so far.
More will come but it's a completely new part that must be designed carefully to avoid problems.

>Knowing what registry.clear() vs registry.clear<Component>() does I thought registry.remove(entity) might do it – turns out it doesn't, it is a no-op instead.

Actually it does not but it may be an interesting feature. Why not?

>What seems to work is registry.destroy(entity); assert(entity == registry.create(entity));, but is that the way it should be done?

When you destroy an entity, all its components are removed **and** its version is increased. Then it enters the list of available entities. It's a LIFO implicit list actually, that's why the same entity is returned if you do a `create` immediately after.

>I can also not remove the feeling of a previous and by now probably ancient version having a registry.clear(entity) for this propose, but I haven't dug into history to checked.

Oh... it did exist? Then is has been removed accidentally. Another good reason to put it back in place.
I didn't even remember it. Usually either I destroy entities or remove specific components.
That's probably why I didn't notice the error. I apologize. Is this the case then?

DonKult:
[...]

PR #450, opened 2020-03-23
Remove all components from an entity if non are given

DonKult:
This PR tries to implement the API as discussed in #446.

Similar to clear() the remove() method is changed to either remove all
specified components or if none are given all components this entity
has. The method also checks (in debug mode) that the components it was
asked to remove are really gone to detect on_destroy listeners re-adding
components.

### DIFF
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -578,16 +578,7 @@ public:
      * @param entity A valid entity identifier.
      */
     void destroy(const entity_type entity) {
-        ENTT_ASSERT(valid(entity));
-
-        for(auto pos = pools.size(); pos; --pos) {
-            if(auto &pdata = pools[pos-1]; pdata.pool->has(entity)) {
-                pdata.remove(*pdata.pool, *this, entity);
-            }
-        }
-
-        // just a way to protect users from listeners that attach components
-        ENTT_ASSERT(orphan(entity));
+        remove(entity);
 
         // lengthens the implicit list of destroyed entities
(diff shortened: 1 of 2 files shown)


src/entt/entity/registry.hpp
```diff
@@ -833,8 +843,12 @@ class basic_registry {
      */
     template<typename... Component, typename It>
     void remove(It first, It last) {
-        ENTT_ASSERT(std::all_of(first, last, [this](const auto entity) { return valid(entity); }));
-        (assure<Component>().remove(*this, first, last), ...);
+        if constexpr(sizeof...(Component) == 0) {
+            std::for_each(first, last, [this](const auto entity) { this->remove(entity); });
```
Kerndog73: Nitpick: I'd use a for loop here to reduce compile time.

DonKult: I see that e62f1edada84f1b029fe5fa1dba303d4ab3d76fb recently replaced std::for_each as I was pretty sure I didn't make that up… I "just" copied and rebased in the wrong order.
```suggestion
            while(first != last) { remove(*(first++)); }
```
No idea of the compile time effects & I see std::for_each used in a few more places (not in that file though). If this (and perhaps other things) are forbidden [function poisoning](https://www.fluentcpp.com/2018/09/04/function-poisoning-in-cpp/) might be a good idea to stop it from reappearing.

skypjack:
I decided to add a new function named `remove_all` since I think `remove(entt)` is ambiguous.

PR #484, opened 2020-05-10
Added `build2` package info

Klaim:
The `entt` packages for `build2` are available there: https://cppget.org/entt
At the moment only `v3.3.0` is avaialble as a "stable" package, `v3.3.1` to `v3.4.0` are currently being tested (see https://queue.cppget.org/entt/ ) and will be available soon.

The sources for the `build2` package are there: https://github.com/build2-packaging/entt

For the moment I didn't made the `build2` package build the tests (only a very minimal test to check that users can use the package, ussing the example code from the readme). I will add `entt`'s original tests later as it is not required for consuming the library.
There is also no build option sets for now, only the default.

Please tell me if this is too much information in the Readme, I can remove some details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,20 @@ Note that benchmarks are not part of this set.
   brew install skypjack/entt/entt
   ```
 
+* [`build2`](https://build2.org), build toolchain for developing and packaging C and C++ code.
+   - Package: [`entt`](https://cppget.org/entt)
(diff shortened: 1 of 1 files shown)


skypjack:
Thank you for the PR. I'm merging it soon.

>Please tell me if this is too much information in the Readme, I can remove some details.

Don't worry. I'll review it a little but it's substantially fine. :+1: :wink:

PR #498, opened 2020-05-26
snapshot: components no longer need to be copyable

alexames:
Replaced calls to std::as_const with std::move so that components can
be moved into place instead of copied when reading them in from a
snapshot.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ skypjack
 
 # Contributors
 
+alexames
 BenediktConze
 bjadamson
 ceeac
--- a/src/entt/entity/snapshot.hpp
+++ b/src/entt/entity/snapshot.hpp
@@ -184,7 +184,7 @@ class basic_snapshot_loader {
                 archive(entt, instance);
                 const auto entity = reg->valid(entt) ? entt : reg->create(entt);
                 ENTT_ASSERT(entity == entt);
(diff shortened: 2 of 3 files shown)


alexames:
It occurs to me that this should probably have some tests to go along with it. I'll get to that in a bit. Adding tests turns out to be nontrivial because the existing tests assume copyability and updating it will take a bit of effort.

skypjack:
>I'll get to that in a bit.

Ok. Ping me if you need help or when you get there. Thanks for contributing.
Also, feel free to add yourself to the list of contributors if you like (file `AUTHORS`). :+1:

alexames:
Sorry for the delay, I have the tests working now. Since I can't store the component directly in the tuple of queues (as that would require a copy) I instead store the data that the non-copyable component contains, so that in can be reconstructed later.

I tried running the tests without the changes to snapshot.hpp, and it fails to build as expected. The new tests only compile and run if the `std::as_const` calls are replaced by `std::move`.

Let me know if there are any changes you'd like to make.

skypjack:
>Sorry for the delay, I have the tests working now.

You've nothing to apologize for. This is an open source project. We do our best in our free time! :wink: 

>Let me know if there are any changes you'd like to make.

Don't worry. I'm merging it on branch `experimental` and making some changes there.
Thank you for pointing this out and for spending your time on it. Really appreciated! :+1:

PR #513, opened 2020-06-19
Non-owning entt::actor

Kerndog73:
Here I have a draft for a non-owning version of `entt::basic_actor` called `entt::basic_handle`. I really had my heart set on the name "handle" but then I realised (well actually the compiler realised 😄) that there's already `entt::handle`. So the alias in the forwarding header is called `handle_grr` because grr naming things is hard! Anyway, we can figure out a name later.

The handle is a reference type while actor is a value type. If actor was `std::vector`, then handle would be `std::span`. This has influenced a number of design decisions. Handles are lightweight types that can be copied and moved around freely. They are intended to be a drop-in replacement for an entity-registry pair. _"having 3 parameters is better than having 6"_.

### Mutability

```C++
entt::registry reg;
entt::entity e = reg.create();

entt::handle h1{e, reg};
const entt::handle h2{e, reg};
entt::const_handle h3{e, reg};
const entt::const_handle h4{e, reg};

h1.assign<int>(); // OK
h2.assign<int>(); // OK
h3.assign<int>(); // Not allowed
h4.assign<int>(); // Not allowed
```

### Implicit conversions

```C++
void add_components(entt::handle);
void get_components(entt::const_handle);

entt::registry reg;
entt::actor actor{reg};
entt::handle handle{actor};

add_components(actor);
get_components(actor);
add_components(handle);
get_components(handle);
```

### Deduction guides

```C++
ENTT_OPAQUE_TYPE(my_entity, entt::id_type);

entt::basic_registry<my_entity> reg;
const auto &const_reg = reg;
my_entity e = reg.create();

entt::basic_actor a1{reg};           // basic_actor<my_entity>
const auto &a2 = a1;
entt::basic_handle h1{a1};           // basic_handle<my_entity>
entt::basic_handle h2{e, reg};       // basic_handle<my_entity>
entt::basic_handle h3{e, const_reg}; // basic_handle<const my_entity>
entt::basic_handle h4{a2};           // basic_handle<const my_entity>
```

I have not attempted to deal with the duplication between the three classes. I wanted to keep it simple and deal with that later. Better to have all this boilerplate in the library than in user code though, right? I was thinking that it might be useful to give these classes a fuller API. Instead of just `assign` and `remove`, there could be `emplace_or_replace`, `remove_if_exists`, `patch`, etc. That is, all (or most) of the registry member functions that operate on a single entity could be added. That will make the duplication much worse but is a separate discussion anyway.

I also haven't added tests for this. I kind of wanted to find a name that doesn't clash (and perhaps discuss the design a bit too) before doing that. I just realised that the examples above are probably halfway towards a test suite.

In a couple of the constructors, I commented `Does this assertion really make sense?`. Would it ever make sense to construct a handle from a valid registry but a null entity? If this class is to be a drop-in replacement for an entity-registry pair then perhaps that assertion should be removed. To be clear, I'm not suggesting that it is removed from actor, just from handle. Or maybe it could be changed to `ENTT_ASSERT(ref.valid(entity) || entity == null)`? Or even `ENTT_ASSERT(reg.valid_or_null(entity))`? I think I remember suggesting `valid_or_null` long ago. Once again, separate discussion!


skypjack:
My main concern is about duplicating the whole API of the registry. This is something to avoid. It's a maintenance burden.
Before discussing anything else, I'd address this point. I can't really figure out an _easy_ way to make it work without duplication at the moment. Any idea?

Innokentiy-Alaytsev:
The simplest from the implementation point of view is having an `operator->()` that return `entt::registry*` and `operator*()` that returns the entity stored inside the `actor`/`handle` so that the final usage is something like:
```c++
auto my_handle = entt::handle {registry.create (), registry};

my_handle->assign (*my_handle, Comp{});
```

Obviously, it comes at the price of more code and possible errors.

Kerndog73:
There's inheritance, sfinae or macros. I don't like the idea of using `operator->` and `operator*`. This is supposed to be a _convenience_ type! I'll see if I can figure something out that isn't too ugly.

Innokentiy-Alaytsev:
I thought about inheritance at first. The problem is that about can't inherit a function and change its signature. So you can't have a special kind of registry that has a defaulted entity parameter it all of its functions. This means that you'll still have to almost copy the API of the registry but with different parameter list. 

I'm not sure how SFINAE can help here. Yes, you can have all kinds of `actor`/`handle` be represented by the same type that is reconfigured depending on whether you want it to be owning or not (that's what I'd have done myself), but the duplication of the registry API is still there.

As for macro, you can easily have it with the said operators. I just don't think it's the best use for macros in this case.

Innokentiy-Alaytsev:
In theory, it could be possible to do all this with minimal duplication, but it would require some templated monstrosities.

1. You'll have to change the registry in such a way that the same internal data can be shared by multiple instances.

2. The registry itself will have to be templated in such a way that there is a variadic template parameter list for each function, dealing with a single entity. Something like `assign (Entity..., Component&&)`. The `Entity...` here is a parameter pack if the type itself, not of the templated function. I'm not sure if it will work in practice, TBH.

3. The registry will have to have a place to store a "default" entity which is used in all the calls.

4. Implementation of each function will have to decide which entity to use: the default one or the one that is passed as a function argument. This will have to be done in an `if constexpr` manner.

5. The actors and handles, in this case, will be a special kind of a registry, that shares the internal state with the registry, provided to the constructor, and defaults the entity to the provided entity.

skypjack:
@Innokentiy-Alaytsev this is a big NO. :smile: 
Before doing all this, I think I can live with `my_handle->assign (*my_handle, Comp{});`.

Innokentiy-Alaytsev:
@skypjack I just was trying to describe a possible solution and all of its complications.

Kerndog73:
Now the registry API is only being duplicated twice instead of three times. I think the best we can do is duplicating it once.

Lawrencemm:
I've personally not seen a convenience handle implemented without duplication unless it's in python or using crazy macros.

At the end of the day, it's not about whether there's duplication, it's about whether there's enough benefit to this to outweigh the maintenance cost.

I've often wished for a convenience wrapper in my code. I'll often to a _lot_ of stuff to just _one_ entity eg. the _character_ entity, and calling registry.get, assign etc. and passing the entity over and over gets tiresome.

So I think if this handle thing can be demonstrated to have general use (I'd love to try it myself to see if it alleviates my pain) then it's worth it, duplication or not.

Kerndog73:
As I already mentioned, I'd much rather have all this boilerplate and duplication in the library than in user code. I'm looking at `basic_actor` and `basic_handle`. I don't see an easy way of dealing with the duplication between them. The way they deal with `const` is different so it's tricky.

Innokentiy-Alaytsev:
It's partially because you use wrong const-qualifier for functions that change state of the registry.

Kerndog73:
I guess we still disagree on that! 😄

Innokentiy-Alaytsev:
I like how @skypjack doesn't say _anything_ about it 😂

skypjack:
Well, the basic question is if we really need both of them.What are the use cases for `actor` and what the ones for `handle`? Can we get around it with a single class that fits all need?

Kerndog73:
They have very different uses. Actor is _"dedicated to those who aren't confident with the entity-component-system architecture"_ and handle is a convenient wrapper around an entity-registry pair. They're just too different. Handle is definitely something I would use (and I've already found a couple of places in EnTT-Pacman where I could use it). Actor on the other hand isn't something I would use. If we were to reduce the maintenance burden by merging the classes, we'd probably end up sacrificing usability somehow. Is the maintenance burden really that bad though?

Lawrencemm:
It seems like actor is just a handle that creates/destroys the entity in the constructor/destructor.

Thus you could extract the rest, put it in the handle class, and leave the auto create/destroy in the actor class and have actor extend from handle.

Kerndog73:
@Lawrencemm It's not quite that simple because of the way `const` works.

Innokentiy-Alaytsev:
I use something like a non-owning handle for simplifying function signatures. I have many cases where I need to work with 2 or 3 entities each of which can belong to a different registry. Having 3 function parameters is way better than having 6. Another feature is creating type-safe function signatures:

```c++
template <
    class TConfigComponent,
    class TQueryRegistry,
    class TConfigRegistry,
    class TDstRegistry,
    class... TConfigIdentifierComponent,
    class... THierarchyComponent,
    class... TIdentifierComponent,
    class... TExcludeComponent >
VoidExpected ApplyConfiguration (
    TypeContainer< TConfigComponent >,
    ConfigIdentifierComponentsList< TConfigIdentifierComponent... >,
    QueryContext< TQueryRegistry > const   i_query,
    ConfigContext< TConfigRegistry > const i_config,
    DstContext< TDstRegistry >             o_dst,
    HierarchyComponentsList< THierarchyComponent... >,
    IdentifierComponentsList< TIdentifierComponent... >,
    ::entt::exclude_t< TExcludeComponent... > = {});
```

Here, `TQueryRegistry`, `TConfigRegistry` and `TDstRegistry` may be different registries of different or the same type or the same registry (depending on the needs of the user code). Passing entities and registries separately is both not fun and not safe. It's very easy to mess up the order or relation between entity and registry. With distinct entity context types (that are more or less the same as the non-owning entity handle) I can guarantee that the user won't mess up the parameters order, and it will be easier to match entities to registries.

Kerndog73:
If difference between the way const works between actor and handle is a little confusing, here's mapping between the handles and some generic type.

```C++
entt::actor               ->  T
entt::actor const         ->  T const
entt::handle              ->  T *
entt::handle const        ->  T * const
entt::const_handle        ->  T const *
entt::const_handle const  ->  T const * const
```

**An actor represents an object. A handle represents a reference to an object.**

Look to the standard library for other types that represent references. A `const std::shared_ptr<T>` let's me modify `T`, a `const std::span<T>` let's me modify `T` and a `const std::vector<T>::iterator` let's me modify `T`. That's why we have `std::shared_ptr<const T>`, `std::span<const T>` and `std::vector<T>::const_iterator`.

Innokentiy-Alaytsev:
As a matter of fact, I think handle should work simpler:
```c++
entt::handle              ->  T *
entt::handle const        ->  T const*
```
But obviously I've lost this argument.

Kerndog73:
@skypjack What do you think so far? I still need to update the wiki. Also might need some more tests. I was thinking about extending the API to bring it inline with the registry by adding functions like `emplace_or_replace`, `remove_if_exists`, `patch`, etc. Another minor thing is the last paragraph of my original comment.

skypjack:
I'm spending some time with my son today, so I will look into this later in the afternoon for a feedback.
However, the first thing that popped to my eyes is that we should give the actor class at least a deprecation cycle before deleting it. :wink:

Lawrencemm:
@Kerndog73 I had an idea for `patch` since I don't like lambdas so much anymore.

Basically `patch` would return a proxy object containing a reference to the component and a reference to registry.

There would be an `operator ->` that delegates to that underlying reference. When this proxy goes out of scope, the registry `patch` is called with a no-op (directly triggering the on_replace would be better here, not sure if possible).

Here's how usage would look:
```
entt::proxy proxy{registry, some_entity);
proxy.patch<some_component>()->some_field = some_different_value; 
// The returned proxy temporary then goes out of scope, triggering the on_replace for the registry.
```
Or:
```
entt::proxy proxy{registry, some_entity);
auto some_component_proxy = proxy.patch<some_component>();
some_component_proxy->some_field = some_different_value;
some_component_proxy->some_other_field = some_other_different_value;
// The returned proxy temporary then goes out of scope, triggering the on_replace for the registry.
```
I think this is much nicer than:
```
entt::proxy proxy{registry, some_entity);
proxy.patch<some_component>([](auto &some_component_instance){
    some_component_instance->some_field = some_different_value;
});
```

Kerndog73:
@Lawrencemm Calling `patch` without a function will trigger the signal. So you could call `get<some_component>()` and then `patch<some_component>()`. Still, `proxy.patch<some_component>()->some_field = some_different_value` is a cool idea. I think something like this would have to be added to the registry first. Then `entt::proxy` will call the registry function. 

BTW, I'm not sold on the name `entt::proxy`. I really liked it yesterday. Your mention of "a proxy object" has made me realise that proxy already means something when you're talking about C++. `entt::handle` is a great name but it's taken 😢 . Open to suggestions.

Lawrencemm:
IMO the current `entt::handle` should be `entt::resource_handle` anyway. Maybe this could be done in a deprecation step or something.

Kerndog73:
@Lawrencemm That's what I said!

Kerndog73:
@skypjack I think this is very close to done. I'm patiently awaiting a review but family comes first obviously. 😉

Innokentiy-Alaytsev:
`avatar`

Innokentiy-Alaytsev:
Or you could wait for `actor` deprecation and use `actor` again.

Kerndog73:
We could deprecate and rename the existing `handle` to `resource_handle`. I know @skypjack didn't like this idea but I think it's acceptable. You can use your IDE to do a find-and-replace for `entt::handle` -> `entt::resource_handle` and you're done in seconds. Because of the deprecation step, and because updating to the next version is optional, I think it's ok.

If we are strongly against that idea then we could also reuse the name `actor` after it's been removed. `actor` is less ambiguous than `proxy` but still not ideal. Maybe I'm fussing over an insignificant detail. 😅

skypjack:
Ok, I looked into this and it looks good. I'd probably make some small changes to the tests but overall I like it.
I don't like much to repeat the API of the registry instead but I don't see any other solution aside what @Innokentiy-Alaytsev suggested. In fact, I'm not against his approach, maybe I even like it more but I see what are the cons of it.

My only concern with renaming deprecating and renaming `handle` at once is that it will break all existing code that uses it.
On the other side, a deprecation step would mean that this class will enter `EnTT` in v3.6 instead of the next one.

Kerndog73:
It doesn't suddenly break. Users have a whole release cycle to deal with the rename. We can also introduce proxy in 3.5 with the warning that it will be renamed.

```c++
// Should also rename cache and loader...
template<typename Resource>
class resource_handle;

template<typename Resource>
using handle [[deprecated("Renamed to resource_handle")]] = resource_handle<Resource>;

template<typename Entity>
class [[deprecated("Will be renamed to handle in the next release")]] proxy;
```

skypjack:
I'm about to merge this. Can I ask you to rollback all the changes related to the `actor` class (doc and deprecation)?
I want it to be in a separate commit and I'm pushing it separately. Thanks.

Kerndog73:
@skypjack Done!

skypjack:
Thank you. In the meantime, I freed the `handle` name for you. :)
I can also make the required changes if you like, let me know.

Kerndog73:
> I can also make the required changes if you like

I'll leave it in your hands. 😉

Innokentiy-Alaytsev:
Please, squash commits before merging. Otherwise it would be incredibly painful to find the origins of the new feature.

skypjack:
Yeah, I'll squash and force push my changes on `experimental`, so as to have @Kerndog73 as author.
Since I'm leaving home for a couple of days, I'm not sure you'll find it upstream before Monday anyway.

Innokentiy-Alaytsev:
I've been thinking about this and came to the conclusion that both of us is right and wrong at the same time. And here's why.

The version of the handle where all the member functions are declared as `const` represents a pointer, as @Kerndog73 described above. 

The version of the handle where only the members that don't change the state of the underlying registry and thus the state of the logical object represented by the entity follows reference semantics.

At first, I was thinking that the reference semantics should be the prefered version. 

Now I think, that both versions should be prefered because they cover different sets of use-cases that intersect to some extent.

The advantage of a pointer-like handle is that it's possible to store a non-constant "pointer" to a constant "object". This handle can be null or be changed during the runtime of the program.

The advantage of a reference-like handle is that the case of function parameters works in a simpler and more familiar way: you can't change the values of the components if the given "reference" is constant. It's not fantastically better and can be implemented with a pointer-like parameter, it's just more familiar and, probably, more idiomatic.

Sorry for this mind dump.

Innokentiy-Alaytsev:
There's a part of the mind dump I forgot to put into the previous message: it would be somewhat nice if we could use names like `pointer` and `reference` for this. This way all the things like `const entt::pointer< entt::entity >`, `entt::pointer< const entt::entity >` and `const entt::reference< entt::entity >` could've looked almost natural.

skypjack:
@Innokentiy-Alaytsev `proxy<const entity>` has the same semantic of what you called a _reference_, am I wrong? I think this is the whole point of @Kerndog73 in one of the replies above.

Innokentiy-Alaytsev:
I slept for 1.5 hours so I can't answer you quickly) I think yes, it does, but I _think_ that the `const proxy< entity >` that does the same thing is somewhat more in like with how you use references in C++. However, the longer I look at it, the more I understand that it's mostly a matter of tastes)

Innokentiy-Alaytsev:
There is one difference:

Imagine that you have two functions: `f_mut(proxy<...>)` that needs to mutate components and `f_pure(proxy<...>)` that accepts the "constant" entity. The `f_mut()` function invokes `f_pure()` function.

If you have the proposed implementation, you'll have to perform a conversion between `proxy<entity>` and `proxy<const entity>` that requires creating a temporary.

If you have a reference-like implementation, you don't need to perform a conversion, because `proxy<entity>` can be used as `const proxy<entity>`. You can even use `const proxy<entity>&` in this case to eliminate yet another copy.

skypjack:
That's true, even though we talk of a copy of two integers in a case that I don't know how common it is.
I think it's really more a matter of tastes, as you correctly pointed out. This one is more _standard-ish_ imho, since the standard library followed this approach in many cases. Moreover, `EnTT` itself already uses this method for views and groups (`view<T>` and `const view<T>` let's you iterate non-const components since `each` is `const` while `view<const T>` is _immutable_ and returns only const instances). Because of this, I think the solution proposed by @Kerndog73 is _better_, not in the sense that the reference-like model is wrong or worse in general but because this one is in line with the overall design of the library.

PR #516, opened 2020-06-27
Properly case entity version for 64-bit support.

Oortonaut:
…ecation messages.

Vulkan.hpp has an annoying enum wrapper class. The changes to page() and offset() are required to select the proper conversion operator.
The changes to registry fixes a type mismatch in 64 bits that could cause unexpected shift results.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,6 +32,7 @@ Milerius
 morbo84
 m-waka
 NixAJ
+Oortonaut
 Paolo-Oliverio
 pgruenbacher
 prowolf
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,6 +17,7 @@ djarek
 DonKult
 drglove
 eliasdaler
+erez-o
 eugeneko
 gale83
 ghost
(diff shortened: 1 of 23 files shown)


Kerndog73:
I can see that the deprecation messages are incorrect.

I can also see how left shifting a 32 bit integer by 32 bits could cause problems (though your fix contains a widening conversion and you forgot the `typename`. I would have done `typename traits_type::entity_type{version}`). I'm also not sure why you changed the indentation.

What I don't understand is the change to `page` and `offset`. You're using a C-style cast and I'm not sure why. It seems unsafe because you won't get a compiler error if `to_integral` returns something that is wider than `std::size_t` (not sure how that could happen) or is signed.

skypjack:
Sorry if I'm late, I was out for vacation.
So, first of all, I'd prefer different PRs for unrelated changes. They are easier to work with, to track and eventually to rollback if needed.
That said, this PR fails to compile on the CI at the moment. Before even going into the details of the individual changes, would it be possible to make it pass all tests on the CI?

Thanks.

Oortonaut:
I'm sorry about pushing two separate changes in one PR.  The changes to sparse_set turned out to be spurious. I corrected the other using KernDog73's suggestion. I've been writing C++ for decades and using entt in a project is one of the ways I'm trying to get up to speed on idiomatic, modern c++.

Also my first Github PR :)

skypjack:
Sure, it's fine, don't worry. Do you also want to add yourself to the list of contributors before merging? See AUTHORS file.

As for the changes, can I ask you to rollback the ones to `meta.hpp`?
I removed those functions on `experimental` branch already since the semantics was different and the deprecation step didn't make much sense. Otherwise, there will be a conflict during the merge.

Thanks.

PR #518, opened 2020-07-01
Deduplicate the STL-compatible container traits

Innokentiy-Alaytsev:


Innokentiy-Alaytsev:
The extracted container traits are there not for code duplication elimination on the EnTT side, but for simplifying new STL-compatibe types registration.

Innokentiy-Alaytsev:
I've split everything in atomic commits for simplifying extracting things that you consider worthy.

I'll rework the traits inheritance so that the type name is not repeated that much.

PR #537, opened 2020-08-05
Ensure context variables types are not empty

dashandslash:
When context variables are empty types, functionalities like `try_ctx` will always return a `nullptr`, which makes impossible to know if the context was set.

### DIFF
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -1599,6 +1599,7 @@ public:
      */
     template<typename Type, typename... Args>
     Type & set(Args &&... args) {
+        static_assert(!std::is_empty<Type>::value);
         unset<Type>();
         vars.push_back(variable_data{type_info<Type>::id(), { new Type{std::forward<Args>(args)...}, [](void *instance) { delete static_cast<Type *>(instance); } }});
         return *static_cast<Type *>(vars.back().value.get());


skypjack:
Wait. Why does it return a nullptr? Context variables are always instantiated. Well, at least, they should be. Am I wrong?

dashandslash:
Well, I had a ctx variable defined as such:
```
struct myCtxVar {};
```
Using `try_ctx<myCtxVar>()` was returning always a `nullptr` even if the context var was set.

skypjack:
> I guess that is because now empty types are not returned..?

Uhm. I don't think so. Probably either it's a bug or you're doing something wrong.
Have you a small example with which to reproduce the issue, so that I can look into it?

dashandslash:
Alright, after  making some  unit tests the issue appears when using a const registry and getting the const ctx variable like so:
```
struct MyCtx
{};
TEST(ctx_test, context_variable_of_empty_types_always_returns_nullptr)
{
    entt::registry r;
    r.set<MyCtx>();
    const auto nonConst = [](entt::registry &r) { EXPECT_TRUE(r.try_ctx<MyCtx>() != nullptr); };
    const auto singleConst = [](const entt::registry &r) { EXPECT_TRUE(r.try_ctx<MyCtx>() != nullptr); };
    const auto doubleConst = [](const entt::registry &r) { EXPECT_TRUE(r.try_ctx<const MyCtx>() != nullptr); };
    nonConst(r);
    singleConst(r);
    doubleConst(r);    // Fails here
}
```

skypjack:
Thank you, I'll look into it as soon as possible. If I were to find a bug, is it a problem if I ping you for a test once fixed?

skypjack:
Ahaaa... yeah, I can confirm that this doesn't work: `try_ctx<const MyCtx>()`. The functions to work with context variable never decay the types they receive. Since `T` and `const T` generate different type identifiers, it doesn't find the second one.
Not sure if we should consider this a bug or not to be honest.

dashandslash:
Gotcha! I guess thabn I can close this PR.
Sorry for the noise 🙇 , I will refactor my code.

Thanks you!

skypjack:
No noise, you're welcome. 😉

PR #538, opened 2020-08-06
Meta function overload support

Innokentiy-Alaytsev:


Innokentiy-Alaytsev:
The current implementation doesn't work when argument cast is required. I look into it tomorrow or the day after or on the third day.

Innokentiy-Alaytsev:
I've split new changes into several commits to make them easier to spot. When you're ready - tell me which commits to squash into the "main" commit (`[meta] Meta-function overload support`).

src/entt/meta/factory.hpp
```diff
@@ -682,11 +682,26 @@ class meta_factory<Type> {
...
         };
 
-        ENTT_ASSERT(!exists(id, type->func));
         ENTT_ASSERT(!exists(&node, type->func));
+
+        if (exists(id, type->func)) {
+            auto const existing_overload =
+                internal::find_if<&std::decay_t<decltype(*type)>::func>(
+                    [id](const auto* curr) {
+                        return curr && (id == curr->id);
```
skypjack, line 691: As far as I remember, `curr` is always != `nullptr` here.

src/entt/meta/meta.hpp
```diff
@@ -1464,6 +1462,66 @@ class meta_type {
 }
 
 
+[[nodiscard]] inline meta_any meta_func::invoke (meta_handle instance, meta_any* const args, const std::size_t sz) const {
```
skypjack, line 1465: I thought about this change. Imho, `meta_func` isn't the right place for it.

In fact, `meta_func` is a wrapper for a function node. I don't expect it to keep iterating the list of available functions of its type for some reasons. It refers to a node, I expect it to use this node and eventually to fail if I do something wrong (like, passing to it the arguments for a different overload).
This is as if you had a pointer to member for an overloaded function, you passed it the list of arguments for another overload and the language or the compiler decided that the pointer to member function is somehow _resolved_ to something different for you under the hood. A little misleading.

My opinion is that we have now these functions in for the meta type:
```
[[nodiscard]] meta_range<meta_func> func() const ENTT_NOEXCEPT;
[[nodiscard]] meta_func func(const id_type id) const;
```
The second one is redundant, you can reach the same result with a range and a manual search.
On the other side, we could rework the second function as:
```
template<typename... Args>
[[nodiscard]] meta_any invoke(const id_type id, Args &&... args) const;
```
Then, we can resolve the overload during name/params resolution.

If you think a moment to it, `meta_type::construct` does exactly the same thing.
Similarly, we could remove the second overload for `data` and introduce a pair `set`/`get` directly on the meta type. And so on for the other meta objects.

This would make using the meta type _easier_too. From:
```
auto res = type.func("foo"_hs).invoke({}, 42);
```
To:
```
auto res = type.invoke("foo"_hs, {}, 42);
```

What are your thoughts on this?
I don't expect you to do that, don't worry. I can easily rework the meta type and prepare the whole thing for the function overload. There shouldn't even be conflicts.

src/entt/meta/factory.hpp
Innokentiy-Alaytsev: I'll check later - you're most likely right. Checking pointers for `nullptr` is just a habbit of mine.

Innokentiy-Alaytsev:
> I thought about this change. Imho, meta_func isn't the right place for it.

I think you're right here. If the `meta_func` represents the "function pointer" concept then iterating the overloads is out of its scope.

> My opinion is that we have now these functions in for the meta type: ...

Isn't the second function exactly the same as the current `meta_type.func()`? Something like this could be more useful:
```c++
template< class... TArg >
[[nodiscard]] meta_func func (id_type const id) const;
```
This function would return an overload accepting the given argument types, one way or the other. A version of this function working with `meta_type` objects could also exist. It's useless when you know all the types but in generic code working with reflection, it would be quite useful. I'd keep both functions, even though the former can be implemented in the terms of the latter.

> On the other side, we could rework the second function as... Then, we can resolve the overload during name/params resolution.

Yes, this version will be more logical. Same goes for `data` and `set`/`get` pair.

Could we go further and replicate the meta-data and meta-function access functions in `meta_any`, especially for those meta-functions that require an instance? This could further ease using `meta`. With what we have now it requires too many hoops to get what you want. I know that it's not that much - just another variable or something like `any.type().func("id"_hs).invoke(any, ...)`, but being able to just `any.invoke("id"_hs, ...)` looks a bit nicer to me. It's shouldn't be the only way to have purely "static" meta-functions.

Innokentiy-Alaytsev:
As for this PR, we could 
1. either defer this feature, rework the way meta-members are accessed and then implement it as part of the update to `meta` - this should be pretty straight forward, like moving the code to another function almost without changes 
2. or introduce it "as is" to get more experience with it and then rework the way meta-members are accessed.

skypjack:
Oh, well, next week I'll take a break during which I will struggle to keep up with my work, then I'll come back stronger than before to fix things as said and proceed with this issue. I'd avoid doing the job twice though. The other changes should be easy at a first glance.

Innokentiy-Alaytsev:
I'll squash commits then and lay the PR on ice until you're ready to use it.

src/entt/meta/meta.hpp
Innokentiy-Alaytsev: For future generations: the shortcuts described here implemented in #552

PR #551, opened 2020-09-08
Truly invalid entt::handle

Vennor:
Following the brief discussion on Discord after my initial impressions of `entt::handle` that were introduced in #513 I'm creating a pull request with proposed changes.

This is a draft and the code will probably change after discussions or will be rejected, so I only changed tests to compile with the current code but didn't add new ones yet or fix the failures. I didn't update the documentation too.

### What's the issue with `entt::handle` in its current form?

It's not possible to store a default initialized and truly invalid handle that doesn't point to any registry and entity.

### Why is that needed?

In my case there are a lot of types that live outside the ECS and point to certain entities in order to inspect them or operate on them. Not pointing to any entity is also a valid case though. Current `entt::handle` makes the described approach much less elegant since it has to be initialized with at least a valid `entt::registry` instance.

### What is the alternative?

I could write my own handle but it would duplicate a major part of `entt::handle` code. I could also wrap it in `std::optional` but that adds unnecessary payload and has a duplicated feature smell to it. Another solution would be to wrap it in a custom type and apply some kind of union hackery to interpret `nullptr` in the first `sizeof(void*)` bytes of the handle as an invalid state.

### What changes?

`entt::handle` takes an `entt::registry` pointer instead of a reference in its constructor. It also has a default constructor. Storing a `nullptr` is fine and part of its functionality now. A few other changes were also done in regards to comparisions, construction and assignments.

### DIFF
--- a/src/entt/entity/handle.hpp
+++ b/src/entt/entity/handle.hpp
@@ -9,18 +9,18 @@ namespace entt {
 
 
 /**
-* @brief Non-owning handle to an entity.
-*
-* Tiny wrapper around a registry and an entity.
-*
-* @tparam Entity A valid entity type (see entt_traits for more details).
-*/
+ * @brief Non-owning handle to an entity.
+ *
+ * Tiny wrapper around a registry and an entity.
+ *
+ * @tparam Entity A valid entity type (see entt_traits for more details).
+ */
 template<typename Entity>
 struct basic_handle {
-    /*! @brief Underlying entity identifier. */
+    / *! @brief Underlying entity identifier.  */
     using entity_type = std::remove_const_t<Entity>;
 
-    /*! @brief Type of registry accepted by the handle. */
+    / *! @brief Type of registry accepted by the handle.  */
     using registry_type = std::conditional_t<
         std::is_const_v<Entity>,
         const basic_registry<entity_type>,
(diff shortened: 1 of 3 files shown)


src/entt/entity/handle.hpp
```diff
@@ -27,17 +27,39 @@ struct basic_handle {
...
+    {}
+
     /**
      * @brief Constructs a handle from a given registry and entity.
      * @param ref An instance of the registry class.
      * @param value An entity identifier.
      */
-    basic_handle(registry_type &ref, entity_type value = null) ENTT_NOEXCEPT
-        : reg{&ref}, entt{value}
+    basic_handle(registry_type *registry, entity_type value = null) ENTT_NOEXCEPT
```
skypjack, line 57: Why a pointer instead of a reference? The goal here is exactly that of having a non-null pointer, that is, a reference.
In case the user has not a registry at hand, another constructor can be used.

src/entt/entity/handle.hpp
```diff
@@ -27,17 +27,39 @@ struct basic_handle {
...
+        : reg{registry}, entt{value}
     {}
 
     /**
-     * @brief Assigns an entity to a handle.
+     * @brief Assigns an entity to a handle while leaving registry unchanged.
```
skypjack, line 62: That's good but... how do I reassign the registry?
As it stands (if I get it right, it's midnight here 🙂), if you create a handle without a registry, there is no way to turn the same in a valid handle. On the other side, if I create a handle with a registry and the null entity, I can make it a valid handle later on. Am I wrong?

src/entt/entity/handle.hpp
```diff
@@ -47,19 +69,64 @@ struct basic_handle {
...
+    }
+
+    /**
+     * @brief Assigns the null object to a handle, invalidating both registry and entity.
      * @return This handle.
      */
     basic_handle & operator=(null_t) ENTT_NOEXCEPT {
-        return (*this = static_cast<entity_type>(null));
+        reg = nullptr;
+        entt = null;
```
skypjack, line 88: Probably this one was wrong already. We should cast the null entity to the underlying type of the entity type, then construct an instance. If it seems tricky, it helps to fully support custom identifiers, also the trickiest ones (and ofc I've realized this when I saw the trickiest ones at work! 😄).

src/entt/entity/handle.hpp
```diff
@@ -47,19 +69,64 @@ struct basic_handle {
...
+    bool operator!=(const basic_handle &other) const ENTT_NOEXCEPT {
+        return !(*this == other);
+    }
+
+    /**
+    * @brief Compares a handle to the null object.
+    * @return True if the handle refers to null registry and null entity, false otherwise.
+    */
+    bool operator==(null_t) const ENTT_NOEXCEPT {
+        return !reg || entt == null;
```
skypjack, line 113: How does the registry matter here? I don't think it's important to be honest.
Comparing with the null entity is an obfuscate call to the bool operator at the end of the day. You want to know if the handle is valid.
Probably the right thing would be to remove this operator and make users use `operator bool` for validation.

src/entt/entity/handle.hpp
```diff
@@ -47,19 +69,64 @@ struct basic_handle {
...
+    }
+
+    /**
+    * @brief Compares a handle to the null object.
+    * @return False if the handle refers to null registry and null entity, true otherwise.
+    */
+    bool operator!=(null_t) const ENTT_NOEXCEPT {
```
skypjack, line 120: Same as above.

src/entt/entity/handle.hpp
```diff
@@ -71,19 +138,27 @@ struct basic_handle {
...
 
     /**
-     * @brief Checks if a handle refers to a valid entity or not.
-     * @return True if the handle refers to a valid entity, false otherwise.
-     */
+    * @brief Checks if a handle refers to non-null registry pointer and entity.
+    * @return True if the handle refers to non-null registry and entity, false otherwise.
+    */
     [[nodiscard]] explicit operator bool() const {
+        return (*this != null);
```
skypjack, line 145: Here the registry matters instead. What if the entity is a valid one but the registry points to null? Should we consider the handle valid? I don't think so.

src/entt/entity/handle.hpp
```diff
@@ -71,19 +138,27 @@ struct basic_handle {
...
         return reg->valid(entt);
     }
 
     /**
      * @brief Returns a reference to the underlying registry.
      * @return A reference to the underlying registry.
      */
-    [[nodiscard]] registry_type & registry() const ENTT_NOEXCEPT {
-        return *reg;
+    [[nodiscard]] registry_type * registry() const ENTT_NOEXCEPT {
```
skypjack, line 160: As above, I don't see why a pointer instead of a reference.

skypjack (changes requested):

Kerndog73:
I feel like `entt::handle` is my baby so naturally I'm inclined to give an opinion on this. ;-)

* `entt::null` represents the null entity so it seems odd to use it for a null handle as well. Given that `h` is a `entt::handle`, this assigns the registry and entity to null:
  ```C++
  h = entt::null;
  ```
  ...and this assigns just the entity to null:
  ```C++
  entt::entity e = entt::null;
  h = e;
  ```
  I would expect assigning `null` to assign the entity to null and not touch the registry. To set a handle to null, you can do `h = entt::handle{}`. `h = {}` is ambiguous because of the entity assignment operator so it probably makes sense for there to be a way to set a handle to null without having to spell out the type.
* `operator==(null_t)` also seems like an obfuscated call to `operator bool` (as skypjack pointed out). I would expect this function to check the entity but I can do `h.entity() == entt::null` or `h.valid()`. Speaking of `valid`...
* `valid` will crash if the registry is null. It might make more sense to do `reg && reg->valid(entt)`. `operator bool` and `valid` do very similar things. Does it make sense to have both? I'm actually not sure.
* The copy constructor and copy assignment operator can be `= default`.
* skypjack mentioned that it's not possible to assign just the registry without having to construct a new handle and assign it. One way to handle this is to add another `operator=` but we can solve another problem by doing things differently. If there were setters for the entity and registry, then `h = {}` is no longer ambiguous. So we could do this:
  ```C++
  entt::handle h;
  entt::registry r;
  entt::entity e = r.create();
  h.registry(r);
  h.entity(e);
  h = {};
  ```
* You added `operator==` between handles but you didn't add comparisons between const and non-const handles. I think the implicit conversion from non-const to const will only work half the time (but I haven't tested this).

src/entt/entity/handle.hpp
Vennor: I think that reassigning the registry alone is not something you do often since most of the time it also causes entity reassignment. In such case `handle = {registry, entity}` is fine.

Still, there's nothing that prevents us from adding `operator=(basic_registry&)` or an explicit setter if you think that's desirable.

src/entt/entity/handle.hpp
Vennor: Good point. When you pass a registry you most likely want to pass a valid object.

src/entt/entity/handle.hpp
Vennor: Agreed. Will remove.

src/entt/entity/handle.hpp
Vennor: Just to make sure: this will return `true` only when both registry and entity aren't null. Is this fine or not?

Vennor:
> I feel like `entt::handle` is my baby so naturally I'm inclined to give an opinion on this. ;-)

Appreciated!

>       I would expect assigning `null` to assign the entity to null and not touch the registry. To set a handle to null, you can do `h = entt::handle{}`. `h = {}` is ambiguous because of the entity assignment operator so it probably makes sense for there to be a way to set a handle to null without having to spell out the type.

Agreed. Maybe a `reset` function?

>     * `valid` will crash if the registry is null. It might make more sense to do `reg && reg->valid(entt)`. `operator bool` and `valid` do very similar things. Does it make sense to have both? I'm actually not sure.

That's because it's a wrapper over `basic_registry::valid`, just like nearly all the other named functions of the handle. I'd certainly expect it to crash in such case and, as a user, prefer writing `handle && handle.valid()` instead. How do you feel about that?

>     * The copy constructor and copy assignment operator can be `= default`.

Indeed. Thanks.

>     * skypjack mentioned that it's not possible to assign just the registry without having to construct a new handle and assign it. One way to handle this is to add another `operator=` but we can solve another problem by doing things differently. If there were setters for the entity and registry, then `h = {}` is no longer ambiguous. So we could do this:
>       ```c++
>       entt::handle h;
>       entt::registry r;
>       entt::entity e = r.create();
>       h.registry(r);
>       h.entity(e);
>       h = {};
>       ```

Sounds good to me. How about you @skypjack?

>     * You added `operator==` between handles but you didn't add comparisons between const and non-const handles. I think the implicit conversion from non-const to const will only work half the time (but I haven't tested this).

Right. Will do!

src/entt/entity/handle.hpp
Vennor: Will do!

src/entt/entity/handle.hpp
skypjack: I'd say yes. After all, is a handle without a valid registry or a valid entity still valid? You can't use it, so probably not.

Vennor:
Implemented a major part of the review suggestions. I went ahead and added registry and entity setters suggested by @Kerndog73.

Kerndog73:
- The indentation seems to have changed.
- There aren't unit tests for the comparison operators. I cloned and tested this myself. It turns out that I was wrong about the implicit conversions only working half the time. You only need one `operator==` and one `operator!=`. I commented the extra ones out and tested every combination of const and non-const and it all worked fine.
- The `operator=` for assigning the entity is still there. I only suggested setters as an alternative to `operator=` to allow `handle = {};`. I was wrong about the ambiguity. `handle = {};` chooses the entity assignment operator (for reasons I don't understand) which is a little confusing if you expect it to have the same effect as `handle = entt::handle{};`.
- There aren't tests that use the default constructor.
- The implicit conversion to const handle suggests that a null registry but non-null entity is an invalid state for the handle. There is also no constructor that takes only the entity. It seems reasonable but is this intended? The registry setter takes a pointer so the user can get to this state. Consider this:
  ```C++
  entt::registry registry;
  entt::entity entity = registry.create();
  entt::handle handle{registry, entity};
  handle.registry(nullptr);
  entt::const_handle chandle{handle};
  // chandle.entity() == entt::null
  // handle.entity() == entity
  ```
  If we consider this state to be invalid, then the registry setter should take a reference. Since the setter takes a reference then the getter should probably return a reference too. If we consider the state valid, then there should be an additional constructor and the implicit conversion operator should be tweaked. If you have an entity but no access to the registry then why use a handle at all? For this reason, I'd be leaning towards this state (null registry, non-null entity) being invalid.
- Another thing to think about is whether there should be a registry setter at all. If you change the registry, then you better change the entity too. So it would make sense to construct a new handle if the registry ever needed to be changed. Thinking about this some more, you would usually use different identifiers for different registries. So there would be only one instance of a registry of a given type. Meaning that if the handle contains a valid registry, there would be no reason to assign another valid registry. So the registry setter really makes no sense at all.
- After deciding that a registry setter actually doesn't make sense, the argument that the getter should return a reference because the setter takes a reference becomes invalid. Returning a pointer makes the most sense anyway since the registry can be null now (and it indeed returns a pointer so no change is required).

Who knew adding a default constructor would cause this much discussion! I'd like to hear what @skypjack has to say about this as I feel like I'm talking around in circles.

skypjack:
Well, if all these inconsistencies come from the fact that we can re-assign the entity and the registry to an existing handle, what about removing this option?
When you construct a handle, it becomes immutable. The only way you have to change it is by copy- or move-assigning it a new instance. I know that the final result is the same but it seems to me a little more explicit than assigning the entitiy to a handle when I don't even know if it contains or not a valid registry.

What are the drawbacks in this case?

Vennor:
>     * The indentation seems to have changed.

Indeed. I must've apply formatting with my style turned on.

>     * There aren't unit tests for the comparison operators. I cloned and tested this myself.

>     * There aren't tests that use the default constructor.

As I remarked in the opening post I'm withholding with writing tests until we reach the final design.

> It turns out that I was wrong about the implicit conversions only working half the time. You only need one `operator==` and one `operator!=`. I commented the extra ones out and tested every combination of const and non-const and it all worked fine.

Good to know. Thanks for testing!

>     * The `operator=` for assigning the entity is still there. I only suggested setters as an alternative to `operator=` to allow `handle = {};`. I was wrong about the ambiguity. `handle = {};` chooses the entity assignment operator (for reasons I don't understand) which is a little confusing if you expect it to have the same effect as `handle = entt::handle{};`.

Turns out I forgot to remove it. 😅

>     * Another thing to think about is whether there should be a registry setter at all. If you change the registry, then you better change the entity too. So it would make sense to construct a new handle if the registry ever needed to be changed. Thinking about this some more, you would usually use different identifiers for different registries. So there would be only one instance of a registry of a given type. Meaning that if the handle contains a valid registry, there would be no reason to assign another valid registry. So the registry setter really makes no sense at all.

That's what I suggested in response to one of @skypjack's comments and also think it's a fine idea.

Vennor:
> Well, if all these inconsistencies come from the fact that we can re-assign the entity and the registry to an existing handle, what about removing this option?
> When you construct a handle, it becomes immutable. The only way you have to change it is by copy- or move-assigning it a new instance. I know that the final result is the same but it seems to me a little more explicit than assigning the entitiy to a handle when I don't even know if it contains or not a valid registry.
> 
> What are the drawbacks in this case?

I can't think of any major drawbacks. If registry and entity getters are present then you can use them when needed to construct a new handle.

Innokentiy-Alaytsev:
It seems like with the removal of the direct entity re-assignment this handle will look similar to the `EntityContext` that I use in my experimental code. The only potential addition I have is a constructor of the `EntityContext` from another context and an entity. This way I reduce the amount of code a little bit.

<details>
  <summary>
    That's what I have.
  </summary>

```c++
#define ASCLEPIA_ECS_ENTITY_CONTEXT(i_name)                                    \
	template < class TRegistry >                                               \
	struct i_name : EntityContext< TRegistry > {                               \
		using Base = EntityContext< TRegistry >;                               \
		using Base::Base;                                                      \
	};                                                                         \
                                                                               \
	template <                                                                 \
	    class TEntity,                                                         \
	    class TRegistry,                                                       \
	    class = ::std::enable_if_t<::std::is_same_v<                           \
	        ::std::decay_t< TRegistry >,                                       \
	        ::entt::registry > > >                                             \
	i_name (TEntity&&, TRegistry &&)                                           \
	    ->i_name<::std::remove_reference_t< TRegistry > >;                     \
                                                                               \
	template < class TOtherRegistry >                                          \
	i_name (EntityContext< TOtherRegistry > const&)                            \
	    ->i_name<::std::remove_reference_t< TOtherRegistry > >;                \
                                                                               \
	template < class TEntity, class TOtherRegistry >                           \
	i_name (TEntity&&, EntityContext< TOtherRegistry > const&)                 \
	    ->i_name< TOtherRegistry >;                                            \
                                                                               \
	static_assert (true)


template < class TRegistry >
struct EntityContext {
	using RegistryType = TRegistry;
	using EntityType   = entity_type_t< RegistryType >;
	using entity_type  = EntityType;

	static_assert (
	    ::std::is_same_v<::entt::registry, ::std::decay_t< RegistryType > >,
	    "Registry is not an entt::registry");


	template < class TOtherRegistry_ >
	static constexpr bool IsCompatibleRegistryType ()
	{
		return ::std::is_same_v<
		    ::std::decay_t< RegistryType >,
		    ::std::decay_t< TOtherRegistry_ > >;
	}


	EntityContext () noexcept = delete;

	template <
	    class TEntity_,
	    class TRegistry_,
	    class = ::std::enable_if_t<::std::is_same_v<
	        ::std::decay_t< TRegistry_ >,
	        ::entt::registry > > >
	explicit EntityContext (
	    TEntity_&& i_entity, TRegistry_&& i_registry) noexcept
	    : entity_{::std::forward< TEntity_ > (i_entity)}
	    , registry_{::std::forward< TRegistry_ > (i_registry)}
	{
		static_assert (
		    ::std::is_same_v<
		        ::std::decay_t< TEntity_ >,
		        entity_type_t<::std::decay_t< TRegistry_ > > >,
		    "Incompatible entity and registry types");

		ASCLEPIA_ECS_DEBUG_ASSERT (
		    registry_.valid (entity_), assert::Level::kLogicCheck);
	}

	template < class TOtherRegistry_ >
	explicit EntityContext (
	    EntityContext< TOtherRegistry_ > const& i_other) noexcept
	    : entity_{i_other.entity_}
	    , registry_{i_other.registry_}
	{
	}


	template < class TEntity_, class TOtherRegistry_ >
	explicit EntityContext (
	    TEntity_&&                              i_entity,
	    EntityContext< TOtherRegistry_ > const& i_other) noexcept
	    : entity_{::std::forward< TEntity_ > (i_entity)}
	    , registry_{i_other.registry_}
	{
		static_assert (
		    ::std::is_same_v<
		        ::std::decay_t< TEntity_ >,
		        entity_type_t<::std::decay_t< TOtherRegistry_ > > >,
		    "Incompatible entity and registry types");

		ASCLEPIA_ECS_DEBUG_ASSERT (
		    registry_.valid (entity_), assert::Level::kLogicCheck);
	}


	EntityType Entity () const noexcept
	{
		return entity_;
	}


	RegistryType& Registry () noexcept
	{
		return registry_;
	}


	RegistryType const& Registry () const noexcept
	{
		return registry_;
	}


	EntityType operator* () const noexcept
	{
		return entity_;
	}


	RegistryType* operator-> () noexcept
	{
		return &registry_;
	}


	RegistryType const* operator-> () const noexcept
	{
		return &registry_;
	}


	template < class TOtherRegistry_ >
	auto operator== (EntityContext< TOtherRegistry_ > const& i_rhs) const
	    noexcept -> ::std::
	        enable_if_t< IsCompatibleRegistryType< TOtherRegistry_ > (), bool >
	{
		return (&registry_ == &i_rhs.registry_) && (entity_ == i_rhs.entity_);
	}


private:
	friend struct EntityContext<::std::remove_const_t< RegistryType > >;
	friend struct EntityContext< const RegistryType >;


	EntityType    entity_;
	RegistryType& registry_;
};


template <
    class TEntity,
    class TRegistry,
    class = ::std::enable_if_t<
        ::std::is_same_v<::std::decay_t< TRegistry >, ::entt::registry > > >
EntityContext (TEntity&&, TRegistry &&)
    ->EntityContext<::std::remove_reference_t< TRegistry > >;


template < class TOtherRegistry >
EntityContext (EntityContext< TOtherRegistry > const&)
    ->EntityContext<::std::remove_reference_t< TOtherRegistry > >;


template < class TEntity, class TOtherRegistry >
EntityContext (TEntity&&, EntityContext< TOtherRegistry > const&)
    ->EntityContext< TOtherRegistry >;


```
</details>

skypjack:
@Vennor 

> If registry and entity getters are present then you can use them when needed to construct a new handle.

Yup, ofc, with setters but without getters.

Vennor:
> Yup, ofc, with setters but without getters.

I assume you meant it the other way around. 😄

skypjack:
I think you misread my message. Check it better. :smile:

Vennor:
The branch is updated again with recent suggestions. Are there any unresolved issues still valid after all the changes?

skypjack:
> The branch is updated again with recent suggestions. Are there any unresolved issues still valid after all the changes?

I don't think so and we can still edit it later but let's wait for a feedback from @Kerndog73 before merging. This is his baby after all. 🙂

Vennor:
Sure! I'll add tests and update the documentation when we're finished with changing the code.

Kerndog73:
My silence was due to being happy with the changes. Removing the setters fixed all the inconsistencies. It's looking good! 👍

skypjack:
@Vennor ping me when you think that we can merge this and thank you a lot for your help. 👍 ❤️

Vennor:
@skypjack, it's ready unless you want me to change something in the tests or documentation.

skypjack:
I think that the only _problem_ remained so far is that literally all comments changed indentation. 😄 
If you don't want to spend more time on this, let me know and I can merge it and rollback the comments for my own. Not a problem at all. 😉

skypjack (approved):

Vennor:
I thought I reverted indentation changes in c10ada2774243ef34e70f99092bb8c90ac37469b. Is it still wrong?

skypjack:
Yeah, I gave a quick look at the changed files. This:
```
/**
 *
```
Became this:
```
/**
*
```
Nothing important though, I can give it a pass if you prefer. Only stylistic stuff. 🙂

Vennor:
Ah, right! Fixed. 😃

Innokentiy-Alaytsev:
@skypjack Please, don't forget to squash things during merging.

skypjack:
There is an issue that I'm trying to solve. Long story short, things like `ASSERT_FALSE(chandle1 == chandle2);` don't invoke `operator==`. Instead, it seems they go through `operator entity_type()`. You can observe it by making the operator explicit.
I think the reason is pretty simple. `entity_type` is an alias for `std::remove_const<Entity>`. Therefore, `operator==(const basic_handle<entity_type> &other)` isn't suitable for the comparison.

Kerndog73:
So I guess you need a comparison operator for `Entity` and another one for `std::conditional_t<std::is_const_v<Entity>, entity_type, const Entity>`.

skypjack:
Or just the same comparison operator, though as a template function:
```
template<typename Type>
[[nodiscard]] bool operator==(const basic_handle<Type> &other) const ENTT_NOEXCEPT {
    return reg == other.registry() && entt == other.entity();
}
```

PR #566, opened 2020-10-07
Allow compilation for Windows with GNU-like compilers

jhasse:
Besides MSVC (and compatible compilers like clang-cl) there's also GCC (MinGW) and GNU-like Clang on Windows. Therefore setting command line flags in CMake should check for `MSVC` instead of `$<PLATFORM_ID:Windows>`.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -40,21 +40,23 @@ function(SETUP_TARGET TARGET_NAME)
     set_target_properties(${TARGET_NAME} PROPERTIES CXX_EXTENSIONS OFF)
     target_link_libraries(${TARGET_NAME} PRIVATE EnTT)
 
-    target_compile_options(
-        ${TARGET_NAME}
-        PRIVATE
-            $<$<NOT:$<PLATFORM_ID:Windows>>:-pedantic -fvisibility=hidden -Wall -Wshadow -Wno-deprecated-declarations>
-            $<$<PLATFORM_ID:Windows>:/EHsc /W1 /wd4996 /w14800>
-    )
-
-    target_compile_options(
-        ${TARGET_NAME}
-        PRIVATE
-            $<$<AND:$<CONFIG:Debug>,$<NOT:$<PLATFORM_ID:Windows>>>:-O0 -g>
-            $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-O2>
-            $<$<AND:$<CONFIG:Debug>,$<PLATFORM_ID:Windows>>:/Od>
-            $<$<AND:$<CONFIG:Release>,$<PLATFORM_ID:Windows>>:/O2>
-    )
+    if(MSVC)
+        target_compile_options(
+            ${TARGET_NAME}
+            PRIVATE
+                /EHsc /W1 /wd4996 /w14800
+                $<$<CONFIG:Debug>:/Od>
+                $<$<CONFIG:Release>:/O2>
+        )
+    else()
+        target_compile_options(
+            ${TARGET_NAME}
+            PRIVATE
+                -pedantic -fvisibility=hidden -Wall -Wshadow -Wno-deprecated-declarations
+                $<$<CONFIG:Debug>:-O0 -g>
+                $<$<CONFIG:Release>:-O2>
+        )
+    endif()
 
     target_compile_definitions(${TARGET_NAME} PRIVATE ENTT_STANDALONE ${ARGN})
 endfunction()


skypjack:
To be honest, I don't even know how this passed the tests in the CI, since this line is unbalanced:
```
$<$<CONFIG:Debug>>:/Od>
```
In any case, the changes make it work with GNU-like compilers while break for example the clang-cl backend of VS.

jhasse:
Ah sorry, I was under Linux when testing this so I only tested the non-MSVC if-branch. I've fixed the unbalanced expression.

> In any case, the changes make it work with GNU-like compilers while break for example the clang-cl backend of VS.

Why so? Note that `if(MSVC)` is true for clang-cl, see https://cmake.org/cmake/help/latest/variable/MSVC.html.

skypjack:
> Why so?

I've just tested it on Windows, VS2019, clang-cl and I can reliably say it didn't work. 😄 
The reasons are beyond my ability to understand the idiosyncrasies of cmake.

jhasse:
What settings did you use? clang-cl on GitHub Actions works and I also couldn't reproduce it on Windows 10 with clang-cl and VS2019 using `cmake -Bbuild -GNinja -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl`.

skypjack:
Ah, wait a moment, it's gtest that is complaining. Let me see what's wrong with it.
> clang-cl on GitHub Actions works
I don't trust it at the moment, since it passed also the tests with unbalanced generators! 😄

skypjack:
Interesting. This makes it complain:
```
project(
    ...
    LANGUAGES CXX
)
```
Not a great problem though. It's enough to set the languages to `CXX C`.
Moreover, VS decided that I want to use the ninja generator by default, thus the original error.

That said, so as to avoid the branch, does this work as well with GNU-like compilers?
```
    target_compile_options(
        ${TARGET_NAME}
        PRIVATE
            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-pedantic -fvisibility=hidden -Wall -Wshadow -Wno-deprecated-declarations>
            $<$<CXX_COMPILER_ID:MSVC>:/EHsc /W1 /wd4996 /w14800>
    )

    target_compile_options(
        ${TARGET_NAME}
        PRIVATE
            $<$<AND:$<CONFIG:Debug>,$<NOT:$<CXX_COMPILER_ID:MSVC>>>:-O0 -g>
            $<$<AND:$<CONFIG:Release>,$<NOT:$<CXX_COMPILER_ID:MSVC>>>:-O2>
            $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:MSVC>>:/Od>
            $<$<AND:$<CONFIG:Release>,$<CXX_COMPILER_ID:MSVC>>:/O2>
    )
```
Or even its single statement form:
```
    target_compile_options(
        ${TARGET_NAME}
        PRIVATE
            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-pedantic -fvisibility=hidden -Wall -Wshadow -Wno-deprecated-declarations>
            $<$<CXX_COMPILER_ID:MSVC>:/EHsc /W1 /wd4996 /w14800>
            $<$<AND:$<CONFIG:Debug>,$<NOT:$<CXX_COMPILER_ID:MSVC>>>:-O0 -g>
            $<$<AND:$<CONFIG:Release>,$<NOT:$<CXX_COMPILER_ID:MSVC>>>:-O2>
            $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:MSVC>>:/Od>
            $<$<AND:$<CONFIG:Release>,$<CXX_COMPILER_ID:MSVC>>:/O2>
    )
```
Even though the first one is a little easier to read from my point of view.

jhasse:
> I don't trust it at the moment, since it passed also the tests with unbalanced generators! 😄

I did fail (cmake returned 1) but the build continued (without any flags added). This is a misfeature of PowerShell's default setting: It will continue running if one command fails (see https://github.com/Azure/azure-functions-powershell-worker/issues/364 for example). People have complained about this to GitHub, too (see https://github.community/t/multiline-commands-on-windows-do-not-fail-if-individual-commands-fail/16753/6), but so far the work-around is to use cmd instead of powershell. As it seems you aren't using any powershell feature, I would suggest that.

skypjack:
> I did fail (cmake returned 1) but the build continued (without any flags added). This is a misfeature of PowerShell's default setting: It will continue running if one command fails

Ahaaaaa... I missed this feature. Thanks for explaining it.

jhasse:
> That said, so as to avoid the branch, does this work as well with GNU-like compilers?

It will work with GNU-like compilers, but not with clang-cl, as its COMPILER_ID is **not** MSVC.

skypjack:
Indeed it worked locally with `-G "Visual Studio 16 2019" -A x64  -T "ClangCl"`. Am I missing something?

jhasse:
Check the build output for:

```
clang-cl : warning : unknown argument ignored in clang-cl: '-pedantic' [-Wunknown-argument] [entt\build\test\odr.vcxproj]
clang-cl : warning : unknown argument ignored in clang-cl: '-fvisibility=hidden' [-Wunknown-argument] [entt\build\test\odr.v
cxproj]
```

If I add nonsense in CMakeLists.txt (on the non-MSVC branch), it will fail for me.

PR #574, opened 2020-10-19
Unreal Engine 4 usage instructions

Koward:
This is a simple documentation page that should explain how to use the EnTT library with a Unreal Engine 4 project.


skypjack:
I'll wait a couple of days to see if someone (other than me) wants to review this and to suggest something. Then, I'll merge it upstream.
Really **thank you very much** for your PR. Really appreciated!

PR #589, opened 2020-11-13
Fixed const, non-const types section intro in entity.md

qzole:
I think this is only a typo, order of reference for const and non const types were swapped around in the intro of "Types: const, non-const and all in between" section.


skypjack:
Good catch. Thank you for pointing this out!

skypjack:
Do you want to add yourself to the list of contributors before merging? See file `AUTHORS` (lexical order).

qzole:
Nah, thanks. If and when I contribute more then a typo fix :P

Issue #592, opened 2020-11-19
Default example in README is confusing, should be explicit about emplace and forwardargs


netpoetica:
In the example in the README, you have a struct without a manual constructor - only default constructor - but then when you use `emplace`, you pass in arguments of two floats, which I presume are intended to set `dx` and `dy`:
```
struct velocity {
    float dx;
    float dy;
};

// ...

int main() {
    entt::registry registry;
    // ..
        if(i % 2 == 0) { registry.emplace<velocity>(entity, i * .1f, i * .1f); }
```
In order for this to be valid, shouldn't there be a manual constructor that takes in 2 floats? How would `emplace`'s forwarding of the args actually be utilized properly in this scenario to set properties of the `velocity` instance?

After seeing this, I started to think there is maybe some magic constructor in C++ structs that knows to take any args in a constructor call and assign them like... in order of their declaration in the struct? This seems quite nuts. I did some Googling to make sure and I couldn't find any such behavior, and running

```
new velocity(i *.1f, i *.1f)
```
will yield "no matching constructor" error.

skypjack:
Eheh, yeah, no magic here. EnTT and specifically its default storage knows if a type is an aggregate or not and it constructs instances accordingly. It helps because most of the times my components are aggregates and I don't want to define a constructor for all of them.
So, the example just works because of this. The example you did is wrong instead. Roughly speaking, the default storage doesn't even invoke `new`, but let's suppose this is the case. The right form isn't this one though:
```
new velocity(i *.1f, i *.1f)
```
Instead, this is what it would be:
```
new velocity{i *.1f, i *.1f}
```
As you can see, [it _just works_](https://godbolt.org/z/1E3c53). :tm: 🙂 
Specifically, [here](https://github.com/skypjack/entt/blob/master/src/entt/entity/storage.hpp#L369) is where the magic happens. Nothing special in fact, but still.

netpoetica:
Hi @skypjack , I appreciate you taking the time to explain. I am still, however, a bit concerned about this from a dev accessibility standpoint - in this call:

```
registry.emplace<velocity>(entity, i * .1f, i * .1f);
```

It the case that arguments 2 and 3 (after `entity`) are intended to be the value of `dx` and `dy`. As per std::aggregate docs, it seems this is very possible:

```
#include <iostream>

int main() {
    struct Velocity
    {
        float dx;
        float dy;
    };
    
    Velocity vel{1.9f, 2.7f};
    
[...]

PR #593, opened 2020-11-19
update example to explain aggregate initialization usage, refs #592

netpoetica:


skypjack:
I'd probably add a note to the documentation when it mentions `emplace` (md files) rather than adding a bunch of comments to the example in the README.
The latter is meant to be minimal and to give only a grasp of the API.

netpoetica:
@skypjack I actually think the bigger concern is in the example, because it is very likely to be the first point-of-entry for users of `entt`. They most likely wouldn't look deeper into the documentation until after they have gotten the minimal example running, and if they see this example and think "what exactly is going on here?" they may never get as far as the more details documentation.

Additionally, `emplace` is so core to the library that even for me (as I have been using `entt` a lot recently), I have never even looked at documentation about emplace. After working with the example, I feel I have intuited how create, emplace, and views work - I only visit documentation for more advanced concepts.

Anyway what I am trying to say here is that I believe the problem isn't anything to do with the documentation, just with this example itself. If you disagree with that, then I might recommend we avoid using aggregate initialization at all in the example just to keep it very simple, and *then* maybe add some documentation to emplace about the convenience of relying on aggregate initialization rather than parameterized constructors.

What are your thoughts?

Kerndog73:
In my opinion, adding constructors to the example will make it more complicated, not less. I think the example perfectly illustrates that `emplace` uses aggregate initialization internally. If I was completely new to EnTT and saw this in the example...

```C++
registry.emplace<position>(entity, i * 1.f, i * 1.f);
```

I'd think to myself "oh, emplace must be using aggregate initialization. that's cool!". And it is cool, isn't it? 😄

netpoetica:
@Kerndog73 I appreciate this feedback :) again, I am not sure for the "average" C++ developer how familiar they are with aggregate initialization. I consider this to be a very edge-case feature in C++ that most people - at least, non-professionals - would be unfamiliar with, but again, this is completely anecdotal. I asked a guy at my company who worked with C++ at Morgan Stanley for something like 15 years if he knew what aggregate initialization was and he said he never heard of it. For me, as more of a hobbyist C++ developer who reads a lot about it, uses it a lot, and is not a professional, I have never used it or even read about (and I have read multiple C++ books by Bjarne himself).

C++ is a language where 10% of it covers 99% of problems, 90% of it covers 1% edge cases, and IMHO aggregate initialization is in that 90% feature set. I am definitely not trying to start a war here - I do not represent the C++ community at large! But coming from a world of working with hundreds of open source libraries professionally, reading docs all day, utilizing libraries via documentation and example, I believe that your baseline example for "Getting Started" should not introduce any of the niche language features pros are using. I am not aware of any other languages that have such a feature, and I would consider this feature of be a magic, nice-to-have thing that is a reminder of why we love C++ so much, but not a good "Getting Started" example.

netpoetica:
Also FWIW, I agree - it *is* cool. It is a very cool feature :) I just don't know how many people are really familiar with this feature, and if it is something your goto getting started example should be leaning so heavily on.

Kerndog73:
> I am not aware of any other languages that have such a feature

Well, C has it but I can see where you're coming from. Initialization in C++ is really complicated. The [initialization section](https://en.cppreference.com/w/cpp/language/initialization) on cppreference is huge. Value initialization, direct initialization, copy initialization... it goes on.

netpoetica:
> The latter is meant to be minimal and to give only a grasp of the API.

I think this point begs the question of, do you *want* to push the users to use aggregate initialization with `entt`, or, do you not really care what approach they use? I believe, if my point about "how familiar is the average C++ developer with this feature?" is correct, then `entt` is inheriting the responsibility of educating its users on aggregate initialization. This is the approach I took in the modification of the README by adding a progressive educational lens where we show the base case (parameterized constructor), but then nudge the user to use aggregate initialization as `entt` intends.

So I think if we want to nudge them, then we also inherit responsibility of educating them. If we don't care, then we can simplify the example to go with base case parameterized constructor, and update documentation for `emplace` to encourage reliance on aggregate initialization. I am open to make these modifications, just need some direction.

skypjack:
So, personally I'm more than willing to make this point explicit in the documentation. Though:

> then entt is inheriting the responsibility of educating its users on aggregate initialization

Although I agree, I find that the right place to do it is the manual, not a snippet not too representative on its cover.
I would prefer to keep the README file minimal and stripped of details, as this is not intended to delve into either `EnTT` or C++ tips and tricks.
In a way, it's even better if that snippet intrigues the reader like it did with you. It will prompt to ask or read the manual and we'll save a friendly RTM in future. 😉 

I also understand that this is a questionable decision, but already from this issue different ideas come out and someone has to make a decision sooner or later. 🙂

netpoetica:
@skypjack Let me know if the above latest modifications are more palatable :)

- Remove loaded comments from example
- In example, have one with and one without  aggregate initialization
- Leave single-line comment to call attention to and create curiosity about aggregate initialization
- Update documentation to progressively explain how `emplace` can work with parameterized constructors or aggregate initialization

> In a way, it's even better if that snippet intrigues the reader like it did with you. It will prompt to ask or read the manual and we'll save a friendly RTM in future. 

I agree with this from an educator perspective, but also, I would rather have been being productive instead of investigating mysteries :) in the end, I am happy to have learned something new, and having this conversation with you folks has made it *super* clear for me, so I thank you! But I still contend that if we're going to take that approach, we should inherit the responsibility of being the teacher. I believe this latest pull request achieves that.

skypjack:
Yeah, thank you and sorry for being late, I'm having a flu in the wrong year.
I'll merge this as soon as I'm fine and edit the commit if needed eventually. 👍

PR #598, opened 2020-11-27
destroy function for handles

Vennor:
Not sure if that's exactly what you had in mind when talking about handle invalidation after destruction, so let me know if I should change it to something else. :)


skypjack:
Thank you very much!! ❤️ 
I'm working on `dev` locally and I'd prefer to finish these changes because they are somewhat related to the commits immediately before.
IF you don't mind, I'll merge this as soon as I'm done, so that the commit will sit on top of mine.

Kerndog73:
- You could probably get away with invalidating only the entity and not the registry. `operator bool` still returns false in that case.
- Also, (nitpick) maybe move the `destroy` function above `emplace` to match the order of the registry.
- For completeness sake, (more nitpicking) you could maybe add the other overload of `destroy` that takes a version. Although that overload is rarely used.

skypjack:
Oh, well, if you plan to make another commit, could you also fix the alignment of the doc?
Super irrelevant thing that your compulsive obsessive dev will fix anyway once merged, but if you could fix it before merging... 😄

skypjack:
Ah, in the meantime I messed up your issue by changing its base. 😅 
Sorry, I apologize, my fault. I should do these things at half past midnight.
The goal was to direct it to `experimental`, so that I can merge it right away.
Forgive me. I feel so guilty! 😰

Vennor:
Oh snap, I messed up the alignment again? Will fix everything tomorrow. Thanks for the hints.

Vennor:
Looks like I broke it even more, lol.

skypjack:
No worries. I can merge it on a branch and pick only the relevant changes. Just let me know when you're done. :+1:

Vennor:
I think I am now. I implemented @Kerndog73's suggestions and skipped yours since I didn't find any formatting problems. Wasn't the the original commit you referred to? I fixed it right away via a force push.

skypjack:
Files changed: 100. 😂 
Ok, I **must** provide a stable branch for PRs or stop using all branches as if they were there only for me.
I'll pick the relevant parts from your commit and close the PR. Sorry again for the inconvenience.

PR #599, opened 2020-11-28
Bring back ENTT_DISABLE_ASSERT

xissburg:
This allows EnTT asserts to be disabled via a compiler option by defining ENTT_DISABLE_ASSERT.

### DIFF
--- a/src/entt/config/config.h
+++ b/src/entt/config/config.h
@@ -27,8 +27,12 @@
 
 
 #ifndef ENTT_ASSERT
-#   include <cassert>
-#   define ENTT_ASSERT(condition) assert(condition)
+#   ifndef ENTT_DISABLE_ASSERT
+#       include <cassert>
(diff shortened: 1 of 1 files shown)


src/entt/config/config.h
```diff
@@ -27,8 +27,12 @@
 
 
 #ifndef ENTT_ASSERT
-#   include <cassert>
-#   define ENTT_ASSERT(condition) assert(condition)
+#   ifndef ENTT_DISABLE_ASSERT
```
skypjack, line 30: I'd probably do the opposite and make disabling asserts higher priority.
After all, you redefine `ENTT_ASSERT` in code but you could want to run the application without it for once.
With your changes, you can never disable asserts if you redefined `ENTT_ASSERT` instead. Does it make sense?

skypjack (changes requested):

PR #610, opened 2020-12-09
Add MAIN_PROJECT check for test option

### LINKED ISSUE #609: Including CTest in the top-level CMakeLists.txt sets BUILD_TESTING=ON for parent projects
#### What is the issue you have?
`BUILD_TESTING` is unset in the main project.
skypjack_entt is included in a main project using FetchContent.
`ENTT_BUILD_TESTING=OFF` is set in the main project.
skypjack_entt's top-level CMakeLists.txt includes CTest regardless.
`BUILD_TESTING=ON` is now set in the main project.

#### Please describe the steps to reproduce the issue.
```
message(STATUS "before BUILD_TESTING := ${BUILD_TESTING}") # prints ""
include(FetchContent)
FetchContent_Declare(
  skypjack_entt
  GIT_REPOSITORY https://github.com/skypjack/entt.git
  GIT_TAG master)
set(ENTT_BUILD_TESTING OFF)
FetchContent_MakeAvailable(skypjack_entt)
message(STATUS "after BUILD_TESTING := ${BUILD_TESTING}") # prints ON
```

#### What is the expected behavior?
CTest should only be included if skypjack_entt is the main project.

#### And what is the actual behavior instead?
CTest is included although skypjack_entt is added as subproject.

globberwops:
This resolves #609 by checking if this is the main project in a CMake project tree.
`ENTT_BUILD_TESTING` will only be set to `ON` if this is the main project.
CTest will only be included, if `ENTT_BUILD_TESTING` is set to `ON`.
`BUILD_TESTING` will therefore only be set, if this is the main project.


skypjack:
Hi, since `ENTT_BUILD_TESTING ` is `OFF` by default, wouldn't it be enough to move the `include(CTest)` line within the `if`?

globberwops:
@skypjack Hi! True, for the sole purpose of resolving this issue, moving the CTest inclusion would suffice.
However, I wanted to take the opportunity to bring the main build configuration closer to what I believe to be more common.

skypjack:
It makes sense. Though, if we move the include under the `if`, what does the other check add to the picture?
Not sure if I'm missing something or not, that's why I'm asking, do not misunderstand me. 🙂

globberwops:
I'm glad you are asking, no worries :)
The check will enable testing by default, when people build EnTT as the main project, say for development or installation.
In both cases, enabling testing seems like a good idea to me. In case people want to use EnTT as a subproject with minimal configuration, testing will be disabled by default.
I am aware that this is a matter of taste, so if you would prefer to keep the impact as minimal as possible, I would agree that moving the include would suffice to resolve the issue at hand.

skypjack:
More than a matter of taste, I'm wondering what will happen for vcpkg or conan users for example.
I'm not sure you **do** want to compile the tests when you _install_ `EnTT` in your system for use through a package manager.
That's why I'm arguing that probably enabling testing should be done explicitly. Does it make sense?

skypjack:
I'm merging this but I decided not to enable tests automatically when `EnTT` is the top dir project. I want to avoid all risks. 🙂

globberwops:
@skypjack Sorry for the late reply. You are absolutely right to avoid the risk of changing the defaults.
Conan should not care about testing, since it is usually based on the CMake install command, which does not install tests by default.
I have never used vcpkg, though, so I have no idea what the implications would be.

Thank you for the discussion, and for accepting the PR!

skypjack:
Oh, thank you for helping. To be honest, I've no clue as well if that would be a problem with any package manager, but I wanted to avoid the risks because in doubt. 🙂

PR #622, opened 2020-12-31
Fixed missing prefix "L"

alexgeek:
The given example doesn't compile without the "L" prefix to the string.


skypjack:
Ah, good catch. Thanks!

PR #629, opened 2021-01-05
entt::any hash function

alexander-v:
This change allows to get std::hash of entt::any content.


skypjack:
I'm not that convinced about this. What's the use case?

alexander-v:
I'd like to use `entt::any` as a key in hash map.
For example, I have an asset system, in which assets can have keys of arbitrary type.
A texture may be loaded from file using `FilePath` type key or generated in runtime using `struct PerlinNoise { float width, height; ... };` key.
There can be lots of other use cases for `entt::any` as a hash map key.

skypjack:
Interesting. So, a sort of _erased key_ for which the type doesn't really matter? Never thought about it actually.

alexander-v:
Yes, in the example above the type of key doesn't matter for asset system.

skypjack:
Ok, I like the idea overall, but there some things of the PR that I would change.
For example, I'm not that keen on polluting `std::` directly from within `EnTT`. Probably an ADL based solution would be nicer here, if you get what I mean.

alexander-v:
Okay, I'll fix it a bit later and reopen the request.

skypjack:
Oh, no, I was looking into this yesterday.
I'm working on some new feature for any because of job, so I delayed this PR a little (there were some conflicts to solve unfortunately).
But still, you can reopen it right away. I'm fine with fixing the conflicts and merging it once done with the changes on my side. :+1:

alexander-v:
Ok, I'll reopen it.
I've been planning to fix your comments, but unfortunately have no time right now.

TerensTare:
`std::hash` provides a specialization for pointer types. I think it would make more sense to use `std::hash<void *>` instead of `return 0;` if the type doesn't define a `std::hash` specialization.

WoLfulus:
> I'd like to use `entt::any` as a key in hash map. For example, I have an asset system, in which assets can have keys of arbitrary type. A texture may be loaded from file using `FilePath` type key or generated in runtime using `struct PerlinNoise { float width, height; ... };` key. Assets themselves are stored in hash map and indexed by these keys. There can be lots of other use cases for `entt::any` as a hash map key.

I have the exact same scenario. Really interested in this one.

skypjack:
Ok, I'm slowly addressing all issues and PRs. This PR has some conflicts and I'd like to split it in two parts (`any` and `meta_any`). Gimme the time to figure it out and it will land upstream. 👍

skypjack:
Unfortunately, I had to revert these changes because the whole things doesn't really work as expected in some projects of mine.

Main problem is that `is_std_hashable` can fail miserably (quite easily) and working around it is hard most of the times.
Briefly, imagine a type which is _conditionally hashable_. In this case, `is_std_hashable` returns true but the code doesn't compile because the underlying type isn't really hashable.
It happens with classes like containers, optionals, tuples and so on. That is, quite a lot of types at the end of the day. It's not feasible to write specializations for everything or ask the user to do so.

This makes supporting the feature as a built-in functionality very (**very**) challenging if not impossible as far as I can tell.
In theory, one could introduce a new class (`any_index`?) that stores an `any` and an erased hash function internally, mostly similar to what I did (it would probably use `any_cast` to achieve the same result). This solution wouldn't affect all uses of `entt::any` in a codebase.

@alexander-v @TerensTare @WoLfulus I'm open to PRs in ☝️ this sense if you are willing to contribute. 🙂 
However, I'm closing this PR because implementing it as a built-in feature of `entt::any` isn't viable.

PR #649, opened 2021-01-31
Delete @bazel_tools//src/conditions:windows_msys.

warriorstar-orion:
This condition is no longer available from Bazel 4.0.0, and when included a `WORKSPACE` built under Bazel 4.0.0, the following error occurs:

```
% blaze build @entt//:entt
ERROR: While resolving configuration keys for @entt//:entt: no such target '@bazel_tools//src/conditions:windows_msys': target 'windows_msys' not declared in package 'src/conditions' (did you mean 'windows_msvc'?) defined by /private/var/tmp/_bazel_warriorstar/e141537f37e88264e88943d724bc897e/external/bazel_tools/src/conditions/BUILD
ERROR: Analysis of target '@entt//:entt' failed; build aborted: no such target '@bazel_tools//src/conditions:windows_msys': target 'windows_msys' not declared in package 'src/conditions' (did you mean 'windows_msvc'?) defined by /private/var/tmp/_bazel_warriorstar/e141537f37e88264e88943d724bc897e/external/bazel_tools/src/conditions/BUILD
INFO: Elapsed time: 3.190s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (13 packages loaded, 101 targets configured)
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -9,7 +9,6 @@ cc_library(
     copts = select({
       "@bazel_tools//src/conditions:windows": _msvc_copts,
       "@bazel_tools//src/conditions:windows_msvc": _msvc_copts,
-      "@bazel_tools//src/conditions:windows_msys": _msvc_copts,
       "//conditions:default": _gcc_copts,
     }),
 )


skypjack:
Doesn't it break everything for users that still rely on Bazel v3? Maybe @zaucy can help here?

warriorstar-orion:
I wish I could say definitively one way or the other, but my primary build target is Darwin. My (very sketchy, very poor) understanding is that:

1. The MSYS2 toolchain is only selected by default for certain bash_* rules, not cc_* rules, so this condition may never actually be selected while building EnTT.
2. Failing that, since there is a default condition in the selector, this will be used in the case a Windows (or any other) platform doesn't get selected.

Testing on Windows 10.0.19042, bazel version 3.7.2, seems to build the library fine. I imagine the only users affected would be those building explicitly with MSYS2/Cygwin.

skypjack:
And these users can get around it anyway, right? Or I didn't understand what you said?

zaucy:
The removal shouldn't break everything for people using older versions of bazel. It might under a certain condition though. I will admit I don't know the condition where `windows_msys` comes into play my best guess would be while running bazel under msys2 (which is now deprecated/not recommended.)

Sometime after I wrote that  `copts` select I had a change of opinion after reading the [Abseil FAQ](https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/930fbec75b452af8bb8c796f5bb754e953e29cf5/FAQ.md#what-is-abi-and-why-don_t-you-recommend-using-a-pre_compiled-version-of-abseil) and it might be best to remove the select altogether.

zaucy (approved):

skypjack:
> it might be best to remove the select altogether.

@warriorstar-orion if you also think that we can remove the `select`, please do it. Otherwise I can merge this upstream right away. Thanks.

warriorstar-orion:
I don't think you can get away with removing it completely. GCC and MSVC will still need distinct command line flags provided (specifically the "/" vs "-" in "-std=c++17"). Bazel is in a weird place right now with how it categorizes "platforms" and how those are linked to compiler toolchains. This is a reasonable set up for a project with a relatively simple build process. The "correct" answer here requires a bit more Starlark code than is really worth.

PR #657, opened 2021-02-10
PS4 build fix

Kartiku:
PS4 headers doesn't define `__cpp_lib_launder` which fails to compile with `-Werror`

Thanks for the nice library

### DIFF
--- a/src/entt/core/any.hpp
+++ b/src/entt/core/any.hpp
@@ -73,7 +73,7 @@ class any {
                     break;
                 }
             } else if constexpr(in_situ<Type>) {
-                #if __cpp_lib_launder >= 201606L
+                #if defined(__cpp_lib_launder) && __cpp_lib_launder >= 201606L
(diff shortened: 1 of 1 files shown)


skypjack:
Ahaa PS4 is really a pain for modern C++! 😄 
Afaik, the last update of the toolchain should work. On the other side, I don't see any issue with the change. So... 🤷‍♂️

skypjack:
I can't try this on PS4 at the moment though.
Can you confirm that you tested it and it works as expected also on the target platform?

Kartiku:
I don't have physical access to ps4 to run the unit tests.  I'm getting our build machine to pass

If it helps boost is doing something similar [here](https://github.com/boostorg/beast/blob/develop/include/boost/beast/core/detail/type_traits.hpp#L102)

skypjack:
Ahah well, yeah, it's quite reassuring actually. 🙂 Thanks for the PR.

PR #674, opened 2021-03-09
build system: add pkg-config support

### LINKED ISSUE #673: [Request] pkg-config support
_TL;DR: Please add pkg-config support to make it easier to depend on this library._

Hello,

I'm in the process of packaging EnTT for [GNU Guix](https://guix.gnu.org/), and I got a recipe that builds, passes the tests, and _even_ compiles projects when including the directory where the headers file end up! Here's [the code](https://paste.sr.ht/~brown121407/ea1fca164240c853e725a675dcc9b0a9bd94780e), if you're curious.

Now, that's great, but it means that users will need to add the include directory by hand in their projects. For example, to add EnTT to a Meson project I have to make my `meson.build` file look a bit like this:

```
...
incdir = include_directories('/home/user/.guix-profile/include/entt')
executable(..., include_directories: incdir)
```

(Maybe there are nicer ways to write that path, but you get the idea -- it has to be written manually.)

I don't know much about pkg-config, but I do know that when a project generates a `.pc` that ends up in the right place, build systems can read it and find the compiler options they need to use to include and link that library.

For example, the `fmt` library, which [builds a `.pc`](https://github.com/fmtlib/fmt/blob/6e1fc01752b3b4c0616d61ab642c714cc7a84670/CMakeLists.txt#L290) from [a template file](https://github.com/fmtlib/fmt/blob/master/support/cmake/fmt.pc.in) can be easily referenced by Meson: `dependency('fmt')` or even used in plain old Makefiles by embedding the pkg-config calls in them.

So my request is to that you would think about generating a `.pc` file, so we can include this library easier in our projects.

Thank you.

brown121407:
Closes #673.

This will generate a new file called `entt.pc`, which can be used by pkg-config to provide compiler flags for EnTT.

Installing EnTT like this:
```
$ mkdir build
$ cd build
$ cmake .. -DCMAKE_INSTALL_PREFIX:PATH=~/entt-install
$ make
$ make install
```
will provide the file in `~/entt-install/lib64/pkgconfig/entt.pc`.

If this file can be found in a folder from `PKG_CONFIG_PATH`, then `pkg-config --cflags entt` will provide the include flag for the compiler.
```
$ PKG_CONFIG_PATH=~/entt-install/lib64/pkgconfig pkg-config --cflags entt
-I/home/seamas/entt-install/include
```
This will make it easy for distro maintainers to install the library so build systems (like Meson, for example) can find that EnTT is installed on the system and include it automatically in projects.

Side note: I tested this by packaging this commit for Guix and it works fine. Meson can find EnTT, my project builds and pkg-config works manually too:
```
$ pkg-config --cflags entt
-I/gnu/store/2xwlnq51jfjc0rp89j62926rlvfiy478-entt-v3.6.0-1.ea21a7d/include
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,12 @@ endif()
 target_compile_features(EnTT INTERFACE cxx_std_17)
 
 #
+# Set pkg-config variables
+#
+set(EnTT_PKGCONFIG ${PROJECT_BINARY_DIR}/entt.pc)
+set(EnTT_PKGCONFIG_DESTDIR ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
+
+#
 # Install EnTT
 #
 
@@ -112,6 +118,12 @@ write_basic_package_version_file(
     EnTTConfigVersion.cmake
     VERSION ${PROJECT_VERSION}
     COMPATIBILITY AnyNewerVersion
+)
+
+configure_file(
+  "${EnTT_SOURCE_DIR}/cmake/in/entt.pc.in"
+  "${EnTT_PKGCONFIG}"
+  @ONLY
 )
 
 configure_package_config_file(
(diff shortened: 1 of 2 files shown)


brown121407:
I think that would have the potential to override the user's option. I haven't seen things in the wild yet that do that. I searched the libs in my user profile and none of them do that. The only thing I could find is this, which also seems to encourage not setting the standard in pkg-config files: https://github.com/NixOS/nixpkgs/issues/33534

PR #728, opened 2021-07-04
fix warning for discarding a nodiscard

Husenap:
In MSVC the `std::array::operator[]` is `[[nodiscard]]` and the return value is discarded in `snapshot.hpp`

I only got this warning when using MSVC.
Compilers tested:
```
MSVC 19.29.30038.1
gcc-9.3
```

### DIFF
--- a/src/entt/entity/snapshot.hpp
+++ b/src/entt/entity/snapshot.hpp
@@ -55,7 +55,7 @@ class basic_snapshot {
 
         while(begin != last) {
             const auto entt = *(begin++);
-            ((reg->template all_of<Component>(entt) ? ++size[Index] : 0), ...);
+            ((reg->template all_of<Component>(entt) ? ++size[Index] : 0u), ...);
         }
 
         (get<Component>(archive, size[Index], first, last), ...);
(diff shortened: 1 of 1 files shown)


src/entt/entity/snapshot.hpp
```diff
@@ -55,7 +55,7 @@ class basic_snapshot {
 
         while(begin != last) {
             const auto entt = *(begin++);
-            ((reg->template all_of<Component>(entt) ? ++size[Index] : size[Index]), ...);
+            ((reg->template all_of<Component>(entt) ? ++size[Index] : 0), ...);
```
skypjack, line 58: Good catch! I've never ever hit this warning, really!
What about 0u though? So that we also avoid the integral promotion.

skypjack (changes requested):

src/entt/entity/snapshot.hpp
Husenap: Good idea, I'll fix that!

skypjack (approved):

Issue #719, opened 2021-06-23
entt symbols are not exported with MinGW64


ElDesalmado:
When exporting symbols from an executable with MinGW64 on Windows nothing is exported.  
The main reason is that class functions are implemented within classes, for example:
```
struct ENTT_API type_seq final {
    [[nodiscard]] static id_type next() ENTT_NOEXCEPT {
        static ENTT_MAYBE_ATOMIC(id_type) value{};
        return value++;
    }
};
```
This SO [answer](https://stackoverflow.com/a/1444034/9363996) mentions, that:
> C++ functions implemented inside of a class declaration are automatically defined inline.

So there is an inconsistency (at least in MinGW64):
- for shared libraries the symbols are exported
- for executables the symbols are inlined and not exported 
- for executables and in-class implemented functions that are explicitly marked `dllexport` the symbols are also exported.

 I know that there are workarounds to just compile `entt` as a dll, or maybe set a `noinline` option , etc. But I really would like to avoid that.

For example, I have a core application that loads plugins, and those plugins are using symbols exported from the core executable.

ElDesalmado:
according to mingw documentation regarding `dllexport`: 
> When applied to C++ classes, the attribute marks defined non-inlined member functions and static data members as exports. Static consts initialized in-class are not marked unless they are also defined out-of-class.

It is not clear though what it implies about `non-inlined` functions: whether they are marked `inline` by user or inlined by compiler. 
I tried to compile a dll that includes `entt/entt.h`, but nothing is exported whatsoever.

skypjack:
Unfortunately, this stuff is not regulated by the standard. So, roughly speaking, it's chaos and nothing more. 😄 

> C++ functions implemented inside of a class declaration are automatically defined inline.

That's true. However, the dll export directive should inhibit the inlining for itself. That's how it works with all major compilers.
Out of curiosity, do you pass `ENTT_API_EXPORT` and `ENTT_API_IMPORT` during compilation? Otherwise `ENTT_API` isn't defined and this would explain the inlining. Also, have you tried the `-fvisibility` flag or equivalent with mingw?

ElDesalmado:
I compile core with `ENTT_API_EXPORT ` and plugins with `ENTT_API_IMPORT`. Though the symbols are not exported. If I explicitly mark class methods for `dllexport`, the symbols are exported. Also I opened a [bug report](https://sourceforge.net/p/mingw-w64/bugs/906/).

skypjack:
[...]

PR #729, opened 2021-07-08
719 add plugin tests

ElDesalmado:
#719 
Added test that are expected to reproduce the problems with using `entt` as a shared library with exported symbols.  

## Difference from existing plugin test
New test uses a shared library (*shared_registry*) with exported `entt` symbols and explicit template instantiations. Core application and dynamically loaded plugins link to shared registry and import its symbols.  
Tests use simple Win32/Linux functions to get symbol address from a dynamically loaded plugins. *cr* library within the existing tests seem to lack such functionality. Moreover it fails to compile with MSYS2 Clang64 and MinGW64.

## Purpose of the tests
To expose volatility of a  platform/implementation specific implementation for exporting inlined functions. 

## Tests
There are several tests that are expected to expose different pitfalls such as:  
1. Symbols of inlined functions are not exported by the linker.
2. Symbols of inlined functions are not imported by the linker due to optimization (tests do pass on Release with MSYS2 MinGW64).

Descriptions for tests can be found within the `.cpp` files in the comments.  

1. *compare_ids_within_core*: tests if template `entt::type_seq` uses the same imported `entt::internal::type_seq` symbols.  
  Will *Fail* if `entt::internal::type_seq` was not imported. In that case exported explicitly instantiated template class will use `entt::internal::type_seq` from within the *Shared Registry*, but "local" type will generate its own version of `entt::internal::type_seq` within the executable.  
  Will *Succeed* if both internal and external template instantiations of `entt::type_seq` use the same `ent::internal::type_seq` symbol. 
  Will *Succeed* (falsely) when exported template is not imported, but generated within the exe alongside with local template instances.

2. *compare_ids_across_modules*: tests if template `entt::type_seq` uses the same imported `entt::internal::type_seq` symbols. In this case only unit-local `entt::type_seq` intsances are used. If `entt::internal::type_seq` was imported ids will not clash.
  Will *Succeed* if  `entt::internal::type_seq` was imported.
  Will *Fail* if both executable and plugin use their own versions of `entt::internal::type_seq`.

3. *test_increment*: passes an entity to the plugin to increment its value.
  Suppose to *Fail* if linker has optimized out symbols on import.


skypjack:
There are a couple of problems that make this not mergeable:
* It doesn't even try to follow the code style of the rest of the library
* There is no way `ENTT_API_IMPORT/EXPORT` can work with a plugin, it doesn't depend on EnTT
* You should do not try to import nor use `entt::internal::type_seq` since it's an internal function (and it's not even expected to be exported actually)

That said, I'm not even sure what this tests are meant to prove, other than showing that something that isn't supposed to work isn't really working:
* The first test fails if the library isn't linked. That's true, it's how `ENTT_API_*` directives work.
* All other tests (with plugins) fail as it ought to be since there isn't a centralized generator, so they _work as intended_ in a sense.

Until EnTT internals change to not use the static generator anymore, you can look at how a centralized generator can be used to work across boundaries with plugins and the like ([link](https://github.com/skypjack/entt/tree/master/test/lib/registry_plugin)).
However, I remain in doubt regarding this PR because its intent isn't really clear to me at the moment.

skypjack:
I close this PR as starving. Fwiw the registry class no longer uses `type_seq` across boundaries (branch `dll`).
Therefore, it's likely that your problem is solved. I suggest to give it a try eventually. I'll merge the branch upstream soon.

PR #776, opened 2021-09-04
Attempt to make header files for EnTT show up in IDE

jonesmz:
I use QtCreator when programming on linux. When working with a project that consumes EnTT, if I navigate to the EnTT folder and tell the IDE to expand the EnTT target, it displays none of the headers.

This isn't so much an EnTT bug as it is an annoying cross-complication between CMake, the way CMake does header-only libs, QtCreator, and the specifics of the EnTT CMakeLists.txt file.

This change attempts to properly describe to any IDEs that the EnTT interface library has, as header files, anything under the "src" directory.

Whether a specific IDE will recognize this and properly display the headers for the EnTT target is up to that IDE.

Note that this also will cause a newly created headerfile, or a deleted header file, in the EnTT src/ folder to re-generate CMake, which I think is desirable, but others may not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,10 @@ include(GNUInstallDirs)
 add_library(EnTT INTERFACE)
 add_library(EnTT::EnTT ALIAS EnTT)
 
+file(GLOB_RECURSE EnTT_H_FILES CONFIGURE_DEPENDS $<BUILD_INTERFACE:${EnTT_SOURCE_DIR}/src>/*.h)
+file(GLOB_RECURSE EnTT_HPP_FILES CONFIGURE_DEPENDS $<BUILD_INTERFACE:${EnTT_SOURCE_DIR}/src>/*.hpp)
+target_sources(EnTT INTERFACE "${EnTT_H_FILES}" "${EnTT_HPP_FILES}")
+
 target_include_directories(
     EnTT
     INTERFACE


skypjack:
It's a fair request and filling the interface sources seems like the right way to me.
Though, I don't like much _globbing_ blindly. Among the other problems, it may also be annoying for you as an user when I add new files to the project.
What if I listed them explicitly?

jonesmz:
> Though, I don't like much globbing blindly. Among the other problems, it may also be annoying for you as an user when I add new files to the project.

I have no objection to the list of headers being filled manually. By all means :-)

I just didn't want to ;)

skypjack:
Ahahah yeah, no worries, I can do it. I'm asking to know if there is any drawbacks of which I'm not aware.

jonesmz:
> Ahahah yeah, no worries, I can do it. I'm asking to know if there is any drawbacks of which I'm not aware.

The only drawback is that if you add a new header, and you don't add it to the list in cmake, it's not in the list in cmake.

If you remove a header, cmake will complain (which, arguably, that's good).

I can't think of any other negatives beyond the increased maintainer burden.

skypjack:
Can I ask you to check if branch `build_system` makes files show up in QtCreator?
It literally contains your changes, though with an explicit list for the files prefixed by `$<BUILD_INTERFACE:${EnTT_SOURCE_DIR}/src>` to avoid/suppress warnings from cmake.

Innokentiy-Alaytsev:
_Disclaimer_: It's opinionated.

Globbing is not a good solution for listing the files. As per CMake documentation, if you run CMake and add files later, the tool won't detect them, and thus they won't be added to the list of sources, etc. To be more precise, if you run `cmake --build <dir>` or directly run the underlying build system, the newly added files won't be detected. If you manually re-run `cmake <dir>`, then everything will work just fine. Not a big deal but an inconvenience.

All this comes into play when you're on the developer side of the globbing target. If you use globbing for listing files for the imported target, everything will likely be all right because the imported targets files are rarely changed.

Adding `INTERFACE` sources to the target has a potentially undesired effect: these sources are added to the `SOURCES` property of downstream targets meaning that every target linking to `EnTT::EnTT` will have those. What makes matters worse is that the files are often not grouped in the IDE sources tree (I may be wrong about it, but it seems like people don't like CMake _that_ much).

Few solutions can resolve this issue:

* use `INTERFACE` sources but also set the `FOLDER` property of each file to make the resulting file tree easier to navigate;
* turn `EnTT::EnTT` into a static library without any exported symbols;
* same as previous but with an additional static library wrapper target instead of changing `EnTT::EnTT`.

jonesmz:
For all its goodness as a tool, this cmake documentation is just utterly terrible.

Their disclaimer is actually incorrect. I tested this on a windows machine when I converted a codebase from an in-house build system to cmake this year (2021) The first thing cmake does, with the Ninja generator anyway, is re-runs the globs and automatically detects whether the list of files has changed.

If it doesn't on some particular platform, that's a bug and should be treated as such by cmake. I have no idea why they thought that disclaimer was worth writing in the first place, as the tool should not have ever behaved the way the disclaimer describes, if it ever did.

I have no feedback on the rest of your comment one way or another.

skypjack:
> these sources are added to the SOURCES property of downstream targets meaning that every target linking to EnTT::EnTT will have those

I think this is actually the most annoying part of adding the sources to the default target.
What if we added a custom target similar to [this one](https://github.com/skypjack/entt/blob/master/CMakeLists.txt#L203) instead and let users decide when to attach it to their own?

Issue #782, opened 2021-09-19
Performance regression in basic_registry::ctx()


cschreib:
Hello! I have been working on a project for a while using EnTT 3.4.0, and recently upgraded to EnTT 3.8.1. The code migration was not very difficult (thanks!), however there seem to have been a significant performance regression in the `basic_registry::ctx()` function.

Below are excerpts of a linux `perf` output before and after migration. These were collected in Debug mode with no optimisation to preserve the full call graph, but the relative performance hit from switching to 3.8.1 is about the same (~50% overall increased runtime) in both Debug and Release configurations.

There is quite a lot of changes listed in the release notes, so I can't pin point which could have triggered this. But from the call graph, it seems that `ctx()` is now using `any` under the hood, and that appears to be the main cause of the overhead. Is this an expected regression, i.e., should I accept that `ctx()` is now slower and cache the output rather than query it all the time?

EnTT 3.4.0:
```
--6.78%--entt::basic_registry<entt::entity>::ctx<game::hex_size>
 |    
  --6.71%--entt::basic_registry<entt::entity>::try_ctx<game::hex_size>
   |    
    --5.55%--std::find_if<__gnu_cxx::__normal_iterator<entt::basic_registry<entt::entity>::variable_data const*, std::vector<entt::basic_registry<entt::entity>::variable_data, std::allocator<en>
     |    
      --5.37%--std::__find_if<__gnu_cxx::__normal_iterator<entt::basic_registry<entt::entity>::variable_data const*, std::vector<entt::basic_registry<entt::entity>::variable_data, std:>
       |    
       --4.83%--std::__find_if<__gnu_cxx::__normal_iterator<entt::basic_registry<entt::entity>::variable_data const*, std::vector<entt::basic_registry<entt::entity>::variable>
        |    
         --3.29%--__gnu_cxx::__ops::_Iter_pred<entt::basic_registry<entt::entity>::try_ctx<game::hex_size>() const::{lambda(auto:1&&)#1}>::operator()<__gnu_cxx::__no>
          |    
           --2.26%--entt::basic_registry<entt::entity>::try_ctx<game::hex_size>() const::{lambda(auto:1&&)#1}::operator()<entt::basic_registry<entt::entity>>
```

EnTT 3.8.1:
```
--21.46%--entt::basic_registry<entt::entity>::ctx<game::hex_size>
 |    
[...]

PR #783, opened 2021-09-19
Add basic_any::is<Type>() and use it to speed up basic_registry::ctx

cschreib:
See ramblings in https://github.com/skypjack/entt/issues/782 for motivation and more information.

### DIFF
--- a/src/entt/core/any.hpp
+++ b/src/entt/core/any.hpp
@@ -252,6 +252,16 @@ public:
     }
 
     /**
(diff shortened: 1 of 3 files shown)


skypjack:
Unfortunately, I cannot accept this PR for a simple reason: comparing vtable's address doesn't work across boundaries.
However, I also have some good news for you. I'll reply directly in the related issue. 👍

PR #798, opened 2021-11-13
document hash-string non-ownership

LeonFretter:
This adds to the documentation of `hashed_string` remarks that no ownership is taken.
I was running into this issue thinking `entt` might take ownership, which is why I want to add these notes.
Not sure about my wording, and in any case happy with alternatives, but some remark that only the passed-in pointer is stored would be great.
I know that `entt` itself doesn't depend on the human-readable string since the value is already computed, but for debugging-purposes one might want to use it, and might get confused by it being undefined.

### DIFF
--- a/src/entt/core/hashed_string.hpp
+++ b/src/entt/core/hashed_string.hpp
@@ -47,6 +47,7 @@ struct fnv1a_traits<std::uint64_t> {
  * counterparts at runtime.<br/>
  * Because of that, a hashed string can also be used in constant expressions if
  * required.
+ * A hashed string only contains a pointer to a user-provided string, and assumes no ownership.
  *
  * @tparam Char Character type.
  */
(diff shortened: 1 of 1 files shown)


skypjack:
Hi, thanks for pointing this out. Do you mind if I change it a little to uniform the approach?
I don't like having the message in the brief, then in the return type, then somewhere else.

LeonFretter:
> Hi, thanks for pointing this out. Do you mind if I change it a little to uniform the approach? I don't like having the message in the brief, then in the return type, then somewhere else.

Hi, thanks for looking into it!
Absolutely no problem, that's what I was hinting at with my remark "happy about alternatives".
See this as an issue with one exemplary approach if you want :wink:

skypjack:
Added your notes as a couple of warnings to make them more visible in the documentation. 👍 
You can find a commit in the `wip` branch, soon upstream on `master`. I close this PR but feel free to continue the discussion here.
Thank you a lot for pointing this out.

Issue #799, opened 2021-11-20
Adding user-definable context variable guards


BrodyHiggerson:
Fitting with the "don't pay for what you don't use" concept, I wanted to have an optional way of injecting some kind of lock/guard around context variables in the registry.

I let the user define how the lock is created since, for me, I use a fiber mutex rather than std::mutex, so I need to be able to define it myself. I'm using the lock in the 'leaf nodes' of the function calls since I can't support recursive mutexes and the end result is the same.

(I'm making this issue since the contribution guidelines don't want immediate PRs. The commit I *would* PR is here: https://github.com/BrodyHiggerson/entt/commit/8870d6237a3229a076c47619ecaa3c1cc7288cd6)

How I'm using it:

```
namespace ftl
{
class Fibtex;
} // namespace ftl

namespace detail
{
ftl::Fibtex& GetContextVarFibtex();
} // namespace detail

#ifndef ENTT_REGISTRY_CONTEXT_GUARD
	#define ENTT_REGISTRY_CONTEXT_GUARD()										\
		ftl::LockWrapper wrapper = rtg::Task::LockFibtex(detail::GetContextVarFibtex());\
		std::lock_guard guard(wrapper);
#endif // #ifndef ENTT_REGISTRY_CONTEXT_GUARD
```

skypjack:
Closing this one, let's continue the discussion in #800 (that is, the PR).👍

Issue #575, opened 2020-10-21
Discussion about potential rename of the context variable functions


s9w:
This was briefly discussed on Discord and the boss agreed on a discussion on this.

The current [context variable](https://entt.docsforge.com/master/entity-component-system/#context-variables) API is
```
registry.set<Username>("s9w");
const auto &username = registry.ctx<Username>();
```

Plus `.try_ctx<T>()` and `.unset<T>()`.

I think the names of these functions is less than optimal for these reasons:
- set and ctx in particular are a pair in function, but not in name.
- `.set()` can be easily mistaken as the counterpart to `.get()`, which it is not.
- `.ctx()` doesn't make it obvious whether it's a function for setting or for retrieving information.

I'm not sure if I have a good replacement. Spontaneously `get_ctx` and `set_ctx` come to mind, or even more explicit `get_ctx_var` and `set_ctx_var`.

Also I'd like to point out the term context variable does have some linguistic overlap with the term [meta context](https://entt.docsforge.com/master/push-entt-across-boundaries/#meta-context).

Thoughts?

skypjack:
I'd avoid the longest version (`*_var`) to be honest. A little too long for my tastes. 🙂

fweig:
Hi everybody, stumbled upon this thread by accident. The context functions were somewhat confusing to me when I first encountered them. So here's my 2 cents:

Just renaming `.set()` to `.set_ctx()` (and `.unset()` to `.unset_ctx()`) might be enough. Makes it clear that `.ctx()` and `.set_ctx()` are a pair but preserves the terse notation for retrieving context variables.

Another advantage: this naming convention is used in the QT framework for getter/setter pairs. So some C++ developers will be already familiar with it.

Green-Sky:
Hi, personal opinion in coming:

I think prefixing `ctx_` is a good idea, because you know with what you are working.

eg:
```
.ctx_set()
.ctx_get()
```

skypjack:
Another option would be to make the context a separate object and have just this:
```
auto &ctx = registry.context();
```
The object can expose the same functionalities of the other containers then, like `emplace` and `remove`.

Innokentiy-Alaytsev:
It this case, the context becomes something like a `handle` for an entity that is always there and can have components assigned to it? Even though it's stupid, it gave me an idea of how entity `0` can be used as `entt::null` almost without wasting space. The almost part, unfortunatelly, suggests that for each context variable there will be a pool with a single instance.

[...]

PR #800, opened 2021-11-20
Added user-defined context variable guard

BrodyHiggerson:
Fitting with the "don't pay for what you don't use" concept, I wanted to have an optional way of injecting some kind of lock/guard around context variables in the registry.

I let the user define how the lock is created since, for me, I use a fiber mutex rather than std::mutex, so I need to be able to define it myself. I'm using the lock in the 'leaf nodes' of the function calls since I can't support recursive mutexes and the end result is the same.

Addresses #799

(Let me know if you want a PR against another branch; maybe main?)

### DIFF
--- a/src/entt/config/config.h
+++ b/src/entt/config/config.h
@@ -69,6 +69,11 @@ static_assert(ENTT_PACKED_PAGE && ((ENTT_PACKED_PAGE & (ENTT_PACKED_PAGE - 1)) =
 #endif
 
 
+#ifndef ENTT_REGISTRY_CONTEXT_GUARD
+	#define ENTT_REGISTRY_CONTEXT_GUARD()
+#endif
+
+
 #ifndef ENTT_STANDARD_CPP
 #    if defined __clang__ || defined __GNUC__
 #       define ENTT_PRETTY_FUNCTION __PRETTY_FUNCTION__
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -1596,6 +1596,7 @@ public:
     template<typename Type, typename... Args>
     Type & set(Args &&... args) {
         unset<Type>();
+        ENTT_REGISTRY_CONTEXT_GUARD()
         vars.emplace_back(std::in_place_type<Type>, std::forward<Args>(args)...);
         return any_cast<Type &>(vars.back());
     }
@@ -1606,6 +1607,7 @@ public:
      */
     template<typename Type>
     void unset() {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         vars.erase(std::remove_if(vars.begin(), vars.end(), [type = type_id<Type>()](auto &&var) { return var.type() == type; }), vars.end());
     }
 
@@ -1634,6 +1636,7 @@ public:
      */
     template<typename Type>
     [[nodiscard]] std::add_const_t<Type> * try_ctx() const {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         auto it = std::find_if(vars.cbegin(), vars.cend(), [type = type_id<Type>()](auto &&var) { return var.type() == type; });
         return it == vars.cend() ? nullptr : any_cast<std::add_const_t<Type>>(&*it);
     }
@@ -1641,6 +1644,7 @@ public:
     /*! @copydoc try_ctx */
     template<typename Type>
     [[nodiscard]] Type * try_ctx() {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         auto it = std::find_if(vars.begin(), vars.end(), [type = type_id<Type>()](auto &&var) { return var.type() == type; });
         return it == vars.end() ? nullptr : any_cast<Type>(&*it);
     }
@@ -1657,6 +1661,7 @@ public:
      */
     template<typename Type>
     [[nodiscard]] std::add_const_t<Type> & ctx() const {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         auto it = std::find_if(vars.cbegin(), vars.cend(), [type = type_id<Type>()](auto &&var) { return var.type() == type; });
         ENTT_ASSERT(it != vars.cend(), "Invalid instance");
         return any_cast<std::add_const_t<Type> &>(*it);
@@ -1665,6 +1670,7 @@ public:
     /*! @copydoc ctx */
     template<typename Type>
     [[nodiscard]] Type & ctx() {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         auto it = std::find_if(vars.begin(), vars.end(), [type = type_id<Type>()](auto &&var) { return var.type() == type; });
         ENTT_ASSERT(it != vars.end(), "Invalid instance");
         return any_cast<Type &>(*it);
@@ -1693,6 +1699,7 @@ public:
      */
     template<typename Func>
     void ctx(Func func) const {
+        ENTT_REGISTRY_CONTEXT_GUARD()
         for(auto pos = vars.size(); pos; --pos) {
             func(vars[pos-1].type());
         }


skypjack:
Unfortunately, this PR is no longer valid (if we consider branch `wip` as it ought to be, that is about to be merged with `master`).
Briefly, I've introduced a `registry_context` object [here](https://github.com/skypjack/entt/blob/wip/src/entt/entity/registry.hpp#L161) to address #575 (that makes perfectly sense to me). All functions from the registry have been moved to this object. [These](https://github.com/skypjack/entt/blob/wip/src/entt/entity/registry.hpp#L1458) are the only functions available in the registry now.

I've no problem merging or reproducing the intent of this PR on a branch, adapting it and then porting it to `master`. There is no need for you to intervene. However, I wanted to ask you what your plans are to avoid conflicts. 🙂

skypjack:
Actually, another option could be to turn the context type in a template parameter.
This way users can inject their own guarded maps or whatever or even _suppress_ it. 🤔

BrodyHiggerson:
@skypjack I actually had to comment out the 

`
#ifndef ENTT_REGISTRY_CONTEXT_GUARD
	#define ENTT_REGISTRY_CONTEXT_GUARD()
#endif
`

section as include ordering was messing with it - some files/libraries would get to config via various utilities and I was struggling to always keep my own header/define before it. I think there's probably a better solution on the cards, but I'd definitely like some way to inject a custom lock either way. A template param could be nice - means the user can typedef entt::registry<ContextType> and use it in their code without worrying about include order shenanigans. I would just be worried about for example having that type _replace_ the entire map and machinery and force the user to re-implement it - I want exactly what the context vars do, just with a lil' interception ;)

TerensTare:
A little late to the party, but I think the best way to solve this is by having a mixin on top of the context. Here's the [best](https://github.com/TerensTare/entt/blob/02c93fbce29b635be652d85299561ff601418e27/src/entt/entity/context.hpp#L139) I could get so far.

The idea is that you create a "locked context" on each thread and use it instead of the regular context. This context would be lock the given mutex on each call to `emplace`/etc. In code it would look something like this:
```cpp
entt::registry registry;

my::mutex ctx_mutex;
entt::locked_context ctx{registry.ctx(), ctx_mutex};

// use ctx just like you would before
```

TerensTare:
Makes sense to me, and looks better than the mixin solution.

skypjack:
Yeah, sure thing. We can also make it a trait actually 🤔 not sure it's worth it though, but I don't like much to dirty the class declaration.

TerensTare:
If this becomes a trait, I think the trait should be passed to the registry as a template parameter. This way you can pick a custom context for every registry, which might be useful for "blueprint registries" (ie. having only blueprint entities). The code would then look something like the following:
```cpp
namespace entt {
    struct default_registry_traits {
        using entity_type = entity;

        template <typename RegistryAlloc>
        using context_type = default_context<RegistryAlloc>;
    };

    template <
        typename Traits, // it's a trait now
        typename Allocator = std::allocator<typename Traits::entity_type>>
    class basic_registry;
}

struct my_traits : entt::default_registry_traits {
    using entity_type = uint64_t;
};
```

The reason why the `Allocator` is not specified in the traits are as following:
1. To break the least amount of code possible. If it were in the traits, the registry would now have one template parameter as opposed to two, which would break code already using a custom allocator.
3. The allocator depends on the entity type, and having to specify it on every specialization can be bug prone. Consider the previous example again:
```cpp
struct my_traits : entt::default_registry_traits {
    using entity_type = uint64_t;
    // oops, using the allocator for entt::entity...
};
```

skypjack:
The other option is to use something on the line of custom data in meta. That is:
```cpp
auto &elem = registry.ctx<my_context_type>();
elem.emplace<my_type>(42, 'c');
```
Where the current type is also the _default_ type. This means that `registry.ctx()` is also backward compatible.
Internally, we only need to allocate the context object and store it with a type id, to debug-assert on use if needed.
This way, one can decide to use a different context type and the whole thing is imho even cleaner than the traits machinery.
Does it make sense? Granted, accessing the context suffers from an extra indirection here, but is it a problem anyway? I don't see the registry context as something used on hot paths in general, it wouldn't make much sense imho.

TerensTare:
I keep saying it again and again, but I like this idea better than the traits. The only thing worth to discuss is the lifetime of the context. Given that now the context type will be dynamic internally, there could be an option to change it at runtime, similar to custom data in meta (eg. the context is initially an `int` variable but then you set it to be a `string`). I would argue against that, as it feels more natural to reason about code when the type doesn't change and there is no benefit to resetting the context to a different type/value (I think).

If one truly needs the feature, it can easily be emulated by making the context represent several states within the type:
```cpp
struct context final
{
    // represent states in one type, instead of having separate types of contexts
    bool is_state_1() const;
    bool is_state_2() const;
    // ...

private:
    // internal data
};
```

However, this means the only way to set the context value (not modify, just set) is when constructing the registry. To me, the simplest way to do this is to have a `basic_any<0u>` parameter to the registry's constructor that defaults to the "default-context" value; I don't see a reason to have a templated constructor instead of `basic_any`.

skypjack:
> there could be an option to change it at runtime

Yes and no actually. Imagine the following:
```cpp
template<typename Type = DefaultType>
auto & ctx() {
    ENTT_ASSERT(context_pair.second == nullptr || context_pair.first == type_id<Type>(), "Mismatch");

    if(context_pair.second == nullptr) {
        context_pair.first = type_id<Type>();
        context_pair.second = std::make_shared<Type>();
    }

    return *static_pointer_cast<Type>(context_pair.second);
}
```
Where `context_pair` is something like `std::pair<type_info, std::shared_ptr<void>>`.
All off the top of my head, I don't think this snippet even compiles, but it should give you an idea of ​​what I mean.
In this case, you can't change the context type at runtime basically, you get an assert and you know it. Once you decide, it's that for the life of the registry. Just... with a lazy type.

TerensTare:
I have a slightly different idea. I'm guessing that `shared_ptr` is used for the type-erased deleter. However, the `shared_ptr` has some extra overhead in memory (allocating the control block, which stores more than just the object). Also, there will only ever be one copy of the context, so there is no need for shared ownership. Also, I would argue that the pair stores the `type_info` even if not used (ie. when assertions are disabled). `type_id` "caches" each `type_info` into static storage, so I think having a reference instead is probably lighter than holding the `type_info` itself. 

Having said that, I think it's best to drop the pair for a `unique_ptr<void, Deleter>`. The deleter doubles up as a way to retrieve the `type_info` of the context when needed, or rather a reference to it. In total, this would need just two pointers to work: the pointer to the context, and the pointer to the deleter, and only one member, the `unique_ptr`. The changes are minimal and the code inside `ctx` is cleaner (subjective).

Here's the [code](https://godbolt.org/z/vKr8Y7vWz).

skypjack:
> The deleter doubles up as a way to retrieve the type_info of the context when needed

This is **REALLY** smart.

skypjack:
Ah! You know why this could be tricky maybe? Because of uses-allocator construction and because we should use the registry allocator to instantiate the context. 🤔

TerensTare:
> This is REALLY smart. 

Thanks, good to hear that. 🙂

> I'd probably use a class with an extra member rather than the same operator() function for everything.

That's fine by me. I only did it like this because it's shorter.

Also probably worth mentioning, but I just remembered that `delete` works with `nullptr` as well (does nothing), so it's perfectly fine if we remove the pointer check on the vtable. I tested with the same code as earlier, and there was no problem.

skypjack:
Yeah, I'm trying to figure out if we can make `allocate_unique` a little more flexible. It doesn't worth it though probably.

TerensTare:
It's probably better to leave `allocate_unique` as is. It shouldn't be too hard to sneak an allocator in the code. [This](https://godbolt.org/z/8vEaqshxP) should be good enough to add allocator support (potentially free for empty allocators), at least for what we discussed so far.

PR #805, opened 2021-12-13
Clarify on_destroy invocations

qwiff-dogg:
It is not obvious that `registry.destroy` does not cause component-level `on_destroy` to be invoked. Hence this addition.


qwiff-dogg:
I see now that `remove_all` is deprecated, so I've closed the PR.

skypjack:
I think the main problem with the review was that `destroy` **does** trigger component-level `on_destroy`. Are you experiencing something different?

qwiff-dogg:
Yes, I'm experiencing something different, but it's probably a bug in my code.

PR #824, opened 2022-01-03
Store size information in hashed string 

StellaSmith:
This patch modifies hashed string to store the size of the given size, or calculate it if a c-string is given, which can be retrieved using the `.size()` member function.

This patch also adds an overload to the `_hs` string literal, allowing it to take a wide string literal and return hashed wide string.

An implicit conversion to `std::string_view` may also be desired, however this patch doesn't include this feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ conan/test_package/build
 .idea
 .vscode
 .vs
+.cache
 CMakeSettings.json
 cpp.hint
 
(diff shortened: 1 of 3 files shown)


skypjack:
I see the idea behind the PR and I kinda agree. However, I've a couple of concerns, mainly due to the fact that this class is largely used already in some projects of mine (or on which I'm working):
* All instances of the class will grow in size. Since they aren't necessarily meant to be used as one-off objects, it may represent a problem in some cases.
* It adds up to the compilation time. It may seem trivial but even a `++` can have an impact when you've tons of hashed strings around.

I'd like to find a way to make the whole thing _optional_. I don't know if there is a way but it would be good/better if the existing code is not affected by this change as we add an extended hashed string alongside the existing one.

PR #833, opened 2022-01-28
meta: reduced size of 3 node structures

morbo84:
These changes reduce, under some quite common circumstances (like `id_type` and `meta_traits` being 4 bytes under an x64 system), the size of `meta_data_node`, `meta_func_node` and `meta_type_node` by 4 bytes each.

### DIFF
--- a/src/entt/meta/factory.hpp
+++ b/src/entt/meta/factory.hpp
@@ -187,11 +187,11 @@ class meta_factory<Type> {
 
         static internal::meta_data_node node{
             {},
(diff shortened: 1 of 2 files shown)


skypjack:
Apparently, many compilers disagree with your asserts. 😅

morbo84:
> Apparently, many compilers disagree with your asserts. 😅

Haha... those bloody asserts weren't supposed to be there 😂 
Let me get rid of them...

PR #844, opened 2022-02-14
enables fetching the number of pending events (size) of dispatcher

WoLfulus:
This enabled calling size() and size<Event>() to dispatcher to check how many pending events there is to be processed on next update.

### DIFF
--- a/src/entt/signal/dispatcher.hpp
+++ b/src/entt/signal/dispatcher.hpp
@@ -263,6 +263,7 @@ public:
      * type.
      *
      * @tparam Event Type of event for which to return the count.
+     * @param id Name used to map the event queue within the dispatcher.
      * @return The total count of pending Event.
      */
     template<typename Event>
--- a/src/entt/signal/dispatcher.hpp
+++ b/src/entt/signal/dispatcher.hpp
@@ -99,7 +99,7 @@ class dispatcher {
 
 public:
     /*! @brief Unsigned integer type. */
-    using size_type = std::size_type;
+    using size_type = std::size_t;
 
     /*! @brief Default constructor. */
     dispatcher() = default;
--- a/src/entt/signal/dispatcher.hpp
+++ b/src/entt/signal/dispatcher.hpp
@@ -34,6 +34,7 @@ class dispatcher {
         virtual void publish() = 0;
         virtual void disconnect(void *) = 0;
         virtual void clear() ENTT_NOEXCEPT = 0;
+        virtual size_t size() = 0;
     };
 
     template<typename Event>
@@ -74,6 +75,10 @@ class dispatcher {
             } else {
                 events.emplace_back(std::forward<Args>(args)...);
             }
+        }
+
+        size_t size() const {
+            return events.size();
         }
 
     private:
@@ -246,6 +251,31 @@ public:
         for(auto &&cpool: pools) {
             cpool.second->publish();
         }
+    }
+
+    /**
+     * @brief Returns the count of pending events for the specified type.
+     *
+     * This method returns the count of pending events for the specified
+     * type.
+     */
+    template<typename Event>
+    size_t size() const {
+        return assure<Event>().size();
+    }
+
+    /**
+     * @brief Returns the total count of pending events.
+     *
+     * This method counts all pending events across all registered event
+     * types.
+     */
+    size_t size() const {
+        size_t events = 0;
+        for(auto &&cpool: pools) {
+            events += cpool.second->size();
+        }
+        return events;
     }
 
 private:
--- a/src/entt/signal/dispatcher.hpp
+++ b/src/entt/signal/dispatcher.hpp
@@ -34,7 +34,7 @@ class dispatcher {
         virtual void publish() = 0;
         virtual void disconnect(void *) = 0;
         virtual void clear() ENTT_NOEXCEPT = 0;
-        virtual size_t size() = 0;
+        virtual std::size_t size() const ENTT_NOEXCEPT = 0;
     };
 
     template<typename Event>
@@ -77,7 +77,7 @@ class dispatcher {
             }
         }
 
-        size_t size() const {
+        std::size_t size() const ENTT_NOEXCEPT {
             return events.size();
         }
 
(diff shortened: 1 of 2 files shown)


src/entt/signal/dispatcher.hpp
```diff
@@ -34,6 +34,7 @@ class dispatcher {
         virtual void publish() = 0;
         virtual void disconnect(void *) = 0;
         virtual void clear() ENTT_NOEXCEPT = 0;
+        virtual size_t size() = 0;
```
skypjack, line 37:
```suggestion
        virtual std::size_t size() const ENTT_NOEXCEPT = 0;
```
I tend to use `std::size_t` rather than `size_t`. Moreover, this should be `const ENTT_NOEXCEPT` probably.

src/entt/signal/dispatcher.hpp
```diff
@@ -76,6 +77,10 @@ class dispatcher {
             }
         }
 
+        size_t size() const {
```
skypjack, line 80:
```suggestion
        std::size_t size() const ENTT_NOEXCEPT {
```

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +253,31 @@ class dispatcher {
         }
     }
 
+    /**
+     * @brief Returns the count of pending events for the specified type.
+     *
```
skypjack, line 258:
```suggestion
     * @tparam Event Type of event for which to return the count.
     * @return The count of pending events for the specified type.
```
Please, remove the comment below too.

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +253,31 @@ class dispatcher {
...
+     * type.
+     */
+    template<typename Event>
+    size_t size() const {
+        return assure<Event>().size();
+    }
+
+    /**
+     * @brief Returns the total count of pending events.
+     *
```
skypjack, line 269:
```suggestion
     * @return The total count of pending events.
```
Please, also remove the comment below.

test/entt/signal/dispatcher.cpp
```diff
@@ -28,6 +28,19 @@ struct receiver {
     int cnt{0};
 };
 
+struct generic_receiver {
```
skypjack, line 31: I don't think we need to add a new class to test the `size` method. A bunch of calls to `size<T>` and `size` here and there should get the job done. It's pretty straightforward after all.

skypjack (changes requested):
I apologize for the delay, I've had some troubles during the last days.
Also, thanks for the PR, it looks fine. I left some comments though.

To be honest, given the size of this PR, you can decide to fix it or I can merge it as is on a branch, fix everything myself and force-push the review on your behalf.
Tell me how what your prefer. 👍

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +253,31 @@ class dispatcher {
...
     }
 
+    /**
+     * @brief Returns the count of pending events for the specified type.
+     *
+     * This method returns the count of pending events for the specified
+     * type.
+     */
+    template<typename Event>
+    size_t size() const {
```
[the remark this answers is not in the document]
WoLfulus: This should also be added to the pools?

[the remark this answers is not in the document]
skypjack: Not necessarily imho. It's internal stuff, we can use `std::size_t` and that's it.

src/entt/signal/dispatcher.hpp
```diff
@@ -76,6 +77,10 @@ class dispatcher {
             }
         }
 
+        std::size_t size() const ENTT_NOEXCEPT {
```
skypjack, line 80:
```suggestion
        std::size_t size() const ENTT_NOEXCEPT override {
```

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +256,37 @@ class dispatcher {
...
     }
 
+    /**
+     * @brief Returns the count of pending events for the specified type.
+     *
+     * This method returns the count of pending events for the specified
+     * type.
+     *
+     * @tparam Event Type of event for which to return the count.
+     * @return The total count of pending Event.
```
skypjack, line 266:
```suggestion
     * @param id Name used to map the event queue within the dispatcher.
     * @return The total count of pending Event.
```

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +256,37 @@ class dispatcher {
...
+     */
+    template<typename Event>
+    size_type size() const ENTT_NOEXCEPT {
```
skypjack, line 269:
```suggestion
    size_type size(const id_type id = type_hash<Event>::value()) const ENTT_NOEXCEPT {
```

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +256,37 @@ class dispatcher {
...
+        return assure<Event>().size();
```
skypjack, line 270:
```suggestion
        return assure<Event>(id).size();
```

src/entt/signal/dispatcher.hpp
```diff
@@ -248,6 +256,37 @@ class dispatcher {
...
+    /**
+     * @brief Returns the total count of pending events.
+     *
+     * This method counts all pending events across all registered event
+     * types.
+     *
+     * @return The total count of pending events.
+     */
+    size_type size() const ENTT_NOEXCEPT {
+        size_type events = 0;
```
skypjack, line 282:
```suggestion
        size_type events{};

```

skypjack (changes requested):

WoLfulus:
I'll take some time to understand the build/test workflow so I can properly test it locally. I was blindly changing the files hoping to see the CI run 🤦🏻‍♂️ . I saw you pushed some more commits, let me know if there's anything I can do here.

skypjack:
No worries at all. If you're fine, I can merge this on a branch, fix the missing const overload of `assure` and push it. Let me know. 👍

PR #864, opened 2022-04-12
Overflow of version and max number of entities

cschreib:
This is a draft PR, since it contains changes that I think would be good to merge, but also other changes that are more explorative at this stage. I have been trying to understand the behavior of EnTT when it comes to reaching the limits of the registry: too many entities, or too many cycles of create/release (overflow of version).

 1. I think I have found an issue in `basic_registry::generate_identifier`, which can return values that cannot be represented in the entity mask, resulting in very odd behavior. I fixed this and added a test.
 2. Trying to understand the logic of version overflow, I added a more extensive test that checks overflow for all entity values. I was surprised by the behavior. First, the registry will prefer to overflow even if there are free entity slots available; I think overflow should be prevented as much as possible. Second, the behavior of overflow depends on which entity is overflowing: in particular, the "last" entity (maximum value of the entity mask) will generate an error on overflow, while the others will overflow silently. I wonder if this is expected. I added the test that shows this, `VersionOverflowAll`, which currently fails on the last iteration of `i`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ ceeac
 ColinH
 corystegel
 Croydon
+cschreib
 cugone
 dbacchet
 dBagrat
--- a/TODO
+++ b/TODO
@@ -18,6 +18,7 @@ WIP:
 * entity-only and exclude-only views
 * custom allocators all over
 * use ENTT_NOEXCEPT_IF as appropriate (ie make compressed_pair conditionally noexcept)
+* add test for maximum number of entities reached
 
 WIP:
 * add user data to type_info
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -301,7 +301,7 @@ class basic_registry {
     }
 
     auto generate_identifier(const std::size_t pos) ENTT_NOEXCEPT {
-        ENTT_ASSERT(pos < entity_traits::to_integral(null), "No entities available");
+        ENTT_ASSERT(pos < entity_traits::to_entity(null), "No entities available");
         return entity_traits::combine(static_cast<typename entity_traits::entity_type>(pos), {});
     }
 


skypjack:
At a first glance, the change to `generate_identifier` looks wrong to me.
I don't know what you mean with _values that cannot be represented in the entity mask_ nor I see in which cases this happens. However, [this line](https://github.com/skypjack/entt/pull/864/files#diff-da5faf7eabeebb6fee126aea280e02f01253d331a019d4dbf08f09e279797e06R305) doesn't work in EnTT. You can't (well, shouldn't) generate the null entity as an user while your change makes it possible for some reasons that I don't fully get.

The changes to `combine` and `construct` might work instead. However, I'd like to know what they fix in practice.
If they are mainly meant for when the entity type is `std::uint8_t`, well, we can safely ignore them and keep the code simple.
You don't want to use something like EnTT if you only have a bunch of elements around, it wouldn't make sense at all.

As for the following:

> Trying to understand the logic of version overflow

I'm more than willing to clear your doubts. 🙂 
Feel free to drop your questions if you like, I'll do my best to answer them. 👍

cschreib:
> At a first glance, the change to generate_identifier looks wrong to me.
I don't know what you mean with values that cannot be represented in the entity mask nor I see in which cases this happens. However, [this line](https://github.com/skypjack/entt/pull/864/files#diff-da5faf7eabeebb6fee126aea280e02f01253d331a019d4dbf08f09e279797e06R305) doesn't work in EnTT. You can't (well, shouldn't) generate the null entity as an user while your change makes it possible for some reasons that I don't fully get.

Ok, I'm working with the following assumptions. Please let me know if any is incorrect :)
 1. The entity "ID" returned by `registry::create()` is composed of two parts: the actual entity, and the version.
 2. The "entity" part of the ID is essentially the "index" of the entity in the various internal containers (for components etc).
 3. This means `registry::create()` must not return two IDs with the same "entity" part (unless the first one has been `release()`d), as otherwise they would share the same components, hence wouldn't be separate entities.
 4. The "version" part of the ID allows internal reuse of allocated memory that has been released, while still differentiating the ID for the "old" entity that used to live at that position in memory, and the ID for the "new" entity that replaced it. From a user perspective, "old" and "new" have different IDs, hence are different entities, even though they share the same space in memory.
 5. The "entity" part and "version" part are both integers, encoded as separate bits of the full entity ID, and extracted with a bitmask. For the `std::uint32_t` entity ID representation, the entity mask is `0x000FFFFF`, and the version mask is `0xFFF00000`.
 6. This means that the maximum number of unique "entity" parts, or memory locations, that the registry can handle in this representation is `1'048'575` (this includes null, I guess, however in my understanding, null is a specific combination of entity part AND version part). 
 7. If I try to go beyond this, e.g., I have an ID equal to `1'048'576`, this actually corresponds to "entity" part equal to `0`, and `version` equal to `1`.
 8. If I have an ID equal to `std::numeric_limits<std::uint32_t>::max()` (which is the null ID), this actually corresponds to "entity" part equal to `1'048'575`, and `version` equal to `4'095`.

If think, if all of this is true, then it's a mistake for `generate_identifier()` to test test against `entity_traits::to_integral(null)`, as this number (equal to `std::numeric_limits<std::uint32_t>::max()`) is going to be larger than `1'048'575`. Hence `registry::create()` could return, say, the ID `1'048'576`, and `emplace_back()` that ID into the `entities` internal array, at index `1'048'576`. But then no one can ever access this data, because all functions (`valid()`, `current()`, etc.) will apply the `to_entity()` function to the returned ID, which will return the index `0`. 

Hence that would make you read data at another location, for another entity. If you're lucky, the entity `0` has a version that is not `1` (or it is released), then you'll get `valid() == false` and an assert will tell you something is wrong. If you're not lucky, the entity `0` exists and has indeed version `1`, and you're now silently and unexpectedly modifying another existing entity.

> The changes to combine and construct might work instead. However, I'd like to know what they fix in practice.
If they are mainly meant for when the entity type is std::uint8_t, well, we can safely ignore them and keep the code simple.
You don't want to use something like EnTT if you only have a bunch of elements around, it wouldn't make sense at all.

Yes, that was done purely to silence narrowing cast warnings when using the `std::uint8_t` entity type. I only needed this for the tests, to simulate running out of entities quicker. That can be done with `std::uint32_t` as well, it will just take longer. I'd expect this warning would also be triggered if using `std::uint16_t`? In my opinion it doesn't make the code more complicated, but it's up to you if you want to keep this or not! I have no use myself for such short data types beyond the testing mentioned earlier.

skypjack:
I see what you mean for the `generate_identifier`. Makes sense. Probably I missed something when I've added pointer stability and thus the tombstones?
In any case, it should be lt-`to_entity(null)` and not lower-or-equal. I still don't get why you want it to be equal since it's not allowed.

---

As a side note, I've enabled the tests here but some of them fail now. 🙂

cschreib:
I have changed `<=` to `<`. I thought the "last" entity was alright, as long as its version wasn't also the "last" version (which then makes it equal to null), but actually this is what was causing the odd failure I reported initially. I removed the additional overflow test, which is no longer necessary.

I also removed the code for the casts to support `std::uint8_t`, and I modified the test to use the standard `entt::entity` instead; it doesn't seem to take too long to run.

skypjack:
Updating the PR to target branch `wip`. Also, the CI isn't too happy yet. 😞

skypjack:
Updating the PR to target branch `wip`. Also, the CI isn't too happy yet. 😞

cschreib:
That would be the test I added, which probably allocates too much memory if `entt::entity` is `std::uint64_t`. It may need to be disabled in that scenario?

skypjack:
I see. My two cents here.
I'm working to rework the entity storage and make it such that all entities < N are considered as _reserved_ when you create entity N and are never used by the registry (unless you ask it to do so explicitly ofc). This would also make writing this test trivial.
Because of that, I would probably fix the issue as you did and only add a line to the todo list to add the test when the rework is upstream.

cschreib:
There you go :)

skypjack:
You can add yourself to the AUTHORS file if you like (lexical order). Otherwise, I can merge this one. 👍

cschreib:
Happy to! Even for one line of code ;P Done.

PR #876, opened 2022-04-24
WIP: apply_fn functional type

DNKpp:
As we already discussed on discord, I would like to see such a helper type in your library. It really makes working with ranges or legacy algorithms more enjoyable.

### DIFF
--- a/src/entt/core/tuple.hpp
+++ b/src/entt/core/tuple.hpp
@@ -5,6 +5,7 @@
 #include <type_traits>
 #include <utility>
 #include "../config/config.h"
+#include "type_traits.hpp"
 
 namespace entt {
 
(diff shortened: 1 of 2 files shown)


skypjack:
I'm likely missing something but how does this:
```cpp
entt::apply_fn{ [](int i, char c) { return i == 42 && c == 'c'; } }(multi);
```
Improve over this?
```cpp
std::apply([](int i, char c) { return i == 42 && c == 'c'; }, multi);
```
🤔

DNKpp:
Well, the first one can be used as an dedicated object, thus it can be used as a functional object for algorithms:
```cpp
entt::registry reg{};
//...
std::ranges::for_each(reg.view<comp1, comp2>().each(), entt::apply_fn{ [](entt::entity e, comp1& c1, comp2& c2) { /*...*/ });
// vs
std::ranges::for_each(reg.view<comp1, comp2>().each(), [](auto&& tuple)
{
    const auto& [e, comp1, comp2] = tuple;
   //...
});```
Just a qol feature.

skypjack:
I see what you mean. Then, since it's literally meant for lambdas, I would probably inherit from `Func` in this case.

skypjack:
Yeah, the coding style is all the way wrong 😅 but I can fix it on my own for such a small PR. Do you want to add yourself to the AUTHORS file before merging?

PR #887, opened 2022-05-10
Add an notice about undefined behaviour in docs.

kcalbCube:


kcalbCube:
Adding event handlers inside event handler invalidates dispatcher::pools. Maybe we should add an notice about that?

skypjack:
Sounds reasonable. Wanna turn this PR into a doc issue? I'm fine with it.

PR #912, opened 2022-07-29
Add a bounded each to observer

cjhowedev:
An observer can now iterate its first N entities with `each`. In the mutable version, this will remove the first N entities from the observer. This allows an observer to be partially iterated when the amount of work is too high for the frame budget.

### DIFF
--- a/src/entt/entity/observer.hpp
+++ b/src/entt/entity/observer.hpp
@@ -422,6 +422,56 @@ public:
         std::as_const(*this).each(std::move(func));
         clear();
     }
+    
+    /**
+     * @brief Iterates the first N entities and applies the given function object to them.
+     *
+     * The function object is invoked for each entity.<br/>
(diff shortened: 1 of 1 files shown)


skypjack:
Technically speaking, new requests should address the `wip` branch and also have some tests to avoid regressions.
That being said, I'm a bit torn about this change. I'd rather expose an iterator or index based access and let the user stop manuallyif required.
The problem with this kind of APIs is that today someone wants an iteration for N element, tomorrow someone asks to add a condition, then a filter, then... and so on. If you give the ability to iterate things externally, filter, stop, whatever, you get rid of all the problems at once and all users have the flexbility they need for their purposes.
Dunno if you get what I mean, just stop me eventually. 👍

skypjack:
Ping.

skypjack:
Closed as not updated/no activity/starving.

PR #924, opened 2022-08-19
Fixed storage component_traits

pkurth:
storage_iterator (storage.hpp) is (or at least might be) templatized using a container (e.g. std::vector) of pointers to components, for example when inserting components into storage via insert().
Therefore, when accessing component_traits, the pointer must be removed (std::remove_pointer_t).

### DIFF
--- a/src/entt/entity/storage.hpp
+++ b/src/entt/entity/storage.hpp
@@ -34,7 +34,7 @@ class storage_iterator final {
 
     using container_type = std::remove_const_t<Container>;
     using alloc_traits = std::allocator_traits<typename container_type::allocator_type>;
-    using comp_traits = component_traits<typename container_type::value_type>;
+    using comp_traits = component_traits<std::remove_pointer_t<typename container_type::value_type>>;
 
     using iterator_traits = std::iterator_traits<std::conditional_t<
         std::is_const_v<Container>,


skypjack:
To be honest, I don't get the use case for this. Can you explain when and why one would need it, please?
Consider that `storage_iterator` is an internal class and it's not meant for the end user. Therefore, I don't fully understand what error of mine these changes would prevent.
Thanks!! 🙂

pkurth:
Sure, sorry if description to my initial PR was too brief.
In my specific use case I copy a bunch of components from one registry to another. For this I employ the insert function. In particular my copy code looks like this:

```
auto v = source.registry.view<component_t>();
auto& s = source.registry.storage<component_t>();
target.registry.insert<component_t>(v.begin(), v.end(), s.cbegin());
```

This internally employs the ```storage_iterator```.
This works fine as is.

However, I also override the ```component_traits``` for some of my components. For example in the following I change the storage page size to a value different than the default 1024:

```
struct example_component
{
	static constexpr std::size_t page_size = 2048;

	/* Component members... */
};
```

This causes a crash in the insert method above, since the ```storage_iterator``` accesses the component traits (in its ```operator[]```), but not those of my component, but those of a _pointer_ to my components. This is because the template argument of ```storage_iterator``` is a ```std::vector<example_component*>```, so a vector of pointers.
Line 37 of ```storage.hpp``` is:

```
using comp_traits = component_traits<typename container_type::value_type>;
```
where ```container_type::value_type``` equals ```example_component*```, **not** ```example_component```.
The iterator thus computes a page_size of 1024 (default page size) and not the 2048 I set above.

I hope this clears things up. If I misuse the API, I apologize :)

Thanks for your amazing library!
\- Philipp

skypjack:
> I also override the component_traits for some of my components

I **DO** love you!! 😍 

I also see your point now, yeah. Thanks for the details.
Really a good catch! It's a shame that the indecent amount of tests didn't get it so far.
Thank you **a lot** for pointing it out. Marking the issue as a bug because, well, that's the real nature of the problem. 🤷‍♂️ 

I think this fix is fine overall but for the fact that I'd like to also add a new test to avoid future regressions.
I'm on vacation at the moment, so you'll have to wait at least next week for that and thus for a merge. Sorry for the delay. 🙂

skypjack:
As a side note:

> I also override the component_traits for some of my components

Not sure you know about it, so I thought it's worth mentioning.
You can add static constexpr tokens to your types and avoid specializations of `component_traits` if you like.

pkurth:
> Not sure you know about it, so I thought it's worth mentioning.
You can add static constexpr tokens to your types and avoid specializations of component_traits if you like.

Yes, that's what I'm doing :) 
From my second post:
```
struct example_component
{
	static constexpr std::size_t page_size = 2048;

	/* Component members... */
};
```

Thanks again for your library. I use it a lot and it is really useful.
\- Philipp

skypjack:
Oh, missed it. Sorry. 👍

PR #930, opened 2022-09-01
Fix/registry member destruction order

DNKpp:
As discussed on discord, I would like to see the context variables get destructed after the component pools. This comes in handy, when a component tears down itself from a context var during destruction. As entt currently doesn't follow a intentional order, I suggest this one to be followed.


skypjack:
Merging on a temporary branch to review a little the test and make it closer to the style of the rest of the library. 🙂

PR #986, opened 2023-02-25
Fixed undefined behavior with sigh publish loop

ChaseCarlson:
It appears that on clang, if you modify calls (i.e. by releasing a connection) from within the scope of the **ranged-based for loop** referencing _calls_, you will get a "vector iterator past end" error.

No error seems to manifest itself on MSVC.

I have changed the **ranged-based for loop** in publish to a **regular for loop**. I have also added a test to check if there is an issue with a disconnect from inside of the loop scope.

This change was not made without cause, as I was experiencing an issue in my project where I had Lua code that disconnected from a signal inside of the callback body of the signal. This code was being run inside of the for loop's body, resulting in a "vector iterator past end" error. This change is necessary to fix my system. I have put a hotfix in but it does not work for my other developers who pull down the project.

### DIFF
--- a/src/entt/signal/sigh.hpp
+++ b/src/entt/signal/sigh.hpp
@@ -168,8 +168,8 @@ public:
      * @param args Arguments to use to invoke listeners.
      */
     void publish(Args... args) const {
-        for(int i=0; i<calls.size(); i++) {
-            auto&& call = calls[i];
+        container_type callsCopy = calls; // Copy allows for calls to be changed inside of the loop without causing any undefined behavior
+        for (auto &&call: callsCopy) {
             call(args...);
         }
     }
--- a/src/entt/signal/sigh.hpp
+++ b/src/entt/signal/sigh.hpp
@@ -168,7 +168,8 @@ public:
      * @param args Arguments to use to invoke listeners.
      */
     void publish(Args... args) const {
-        for(auto &&call: std::as_const(calls)) {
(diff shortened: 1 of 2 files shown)


ChaseCarlson:
I have now changed back to a ranged-based for loop since we are making a copy of the calls container prior to using it to prevent other edge cases from appearing (i.e. the case you mentioned in the Discord help channel). The copied calls container will remain unmutated, which is why I switched back to range-based for loop for readability.

skypjack:
Hello. I'm pretty sure that copying isn't something that everyone wants.
It boils down to allocations and extra deletes for each event, all to support a feature perhaps unwanted.
If you really want it, you can copy the sigh class on your own and invoke `publish` on the copy. 
Otherwise, it's literally a performance penalty for all other users at the moment. 🙂

Modifying a container while you visit it isn't exactly easy to support and more or less all the standard containers reject this possibility.
It's easy to see why as well as find an example that breaks yet another proposed solution.
On the other hand, more often than not it's also easy to get around the problem with a change to the surrounding code.

Can I ask you to describe your problem a little further?

skypjack:
Another way to support your request is to get rid of the `before` functions in the `sink` (which are imho as useless as they can get), then swap-and-pop listeners upon destruction.
In this case, we can just iterate backward within `publish` and get the job done both when removing and adding listeners.

PR #1005, opened 2023-04-13
Implement support for dynamic sized storages

Naios:
Hi @skypjack,

as discussed in Discord I'm proposing the following code to add dynamic (runtime) sized object support to entt.

I'm looking forward to your feedback.

---

## Points to discuss

There is also the possibility to move the methods from the traits objects into the stateful allocator, but I thought it would be much cleaner to keep the responsibility of the allocator and the storage related methods separate.

## ToDo

- [ ] Implement missing tests
- [ ] Review whether it is intended that `move_element` can alias (the element to swap from and to can point to the same object by the storage algorithms): e.g. `[[maybe_unused]] auto unused = std::exchange(elem, std::move(other));` in `basic_storage::pop`.
- [ ] Review the polymorphic allocator workaround
- [x] Discuss heavily copying of allocators in the storage class (before this PR), those should be passed as lvalue reference. Would be the possibility for a follow-up commit.

---

## Usage

```cpp
entt::registry registry;

    // Create a dynamic trait object
    entt::component_traits<dynamic_data> const traits(32);
    ASSERT_EQ(traits.dynamic_size(), 32);

    // Create an allocator which supplies objects of the dynamic size
    dynamic_allocator<dynamic_data> const alloc{traits.dynamic_size()};

    // Create the storage for the dynamic sized data,
    // and assert that it was created with the specified parameters.
    ASSERT_TRUE(registry.try_create_storage<dynamic_data>(alloc, traits));

    auto &storage = registry.storage<dynamic_data>();

    // Ensure that the allocator matches the provided one
    ASSERT_EQ(storage.get_allocator(), alloc);

    // Ensure that the trait matches the provided one
    ASSERT_EQ(storage.get_traits(), traits);

    entt::entity const e1 = registry.create();
    entt::entity const e2 = registry.create();

    {
        dynamic_data &d1 = registry.emplace<dynamic_data>(e1);
        dynamic_data &d2 = registry.emplace<dynamic_data>(e2);
```

### DIFF
--- a/src/entt/core/memory.hpp
+++ b/src/entt/core/memory.hpp
@@ -222,6 +222,16 @@ struct uses_allocator_construction<std::pair<Type, Other>> {
     }
 };
 
+/**
+ * @brief deduces to a true type if the allocator provides a resource method.
+ */
+template<typename Allocator, typename = void>
+struct is_polymorphic_allocator: std::false_type {
+};
+template<typename Allocator>
+struct is_polymorphic_allocator<Allocator, std::void_t<decltype(std::declval<Allocator &>().resource())>>: std::true_type {
+};
+
 } // namespace internal
 
 /**
@@ -280,8 +290,25 @@ constexpr Type make_obj_using_allocator(const Allocator &allocator, Args &&...ar
  * @return A pointer to the newly created object of the given type.
  */
 template<typename Type, typename Allocator, typename... Args>
-constexpr Type *uninitialized_construct_using_allocator(Type *value, const Allocator &allocator, Args &&...args) {
-    return std::apply([value](auto &&...curr) { return new(value) Type(std::forward<decltype(curr)>(curr)...); }, internal::uses_allocator_construction<Type>::args(allocator, std::forward<Args>(args)...));
+constexpr Type *uninitialized_construct_using_allocator(Type *value, Allocator &allocator, Args &&...args) {
+    return std::apply([value, &allocator](auto &&...curr) mutable -> Type * {
+        using alloc_type = std::decay_t<Allocator>;
+        using alloc_traits = std::allocator_traits<alloc_type>;
+
+        static_assert(std::is_convertible_v<Type *, typename alloc_traits::value_type *>,
+                      "Expected the allocator to allocate elements convertible to type Type!");
+
+        // Workaround for polymorphic allocator not propagating this prio to C++20
+        // when calling std::polymorphic_allocator::construct (at least on MSVC).
+        // This workaround can be removed once entt requires C++20.
+        if constexpr(internal::is_polymorphic_allocator<alloc_type>::value) {
+            new(value) Type(std::forward<decltype(curr)>(curr)...);
+        } else {
+            alloc_traits::construct(allocator, value, std::forward<decltype(curr)>(curr)...);
+        }
+
+        return value;
+    }, internal::uses_allocator_construction<Type>::args(allocator, std::forward<Args>(args)...));
 }
 
 } // namespace entt
--- a/src/entt/entity/mixin.hpp
+++ b/src/entt/entity/mixin.hpp
@@ -77,6 +77,8 @@ class sigh_mixin final: public Type {
 public:
     /*! @brief Allocator type. */
     using allocator_type = typename Type::allocator_type;
+    /*! @brief Component traits. */
+    using traits_type = typename Type::traits_type;
     /*! @brief Underlying entity identifier. */
     using entity_type = typename Type::entity_type;
     /*! @brief Expected registry type. */
@@ -89,9 +91,10 @@ public:
     /**
      * @brief Constructs an empty storage with a given allocator.
      * @param allocator The allocator to use.
+     * @param traits The type_traits to use.
      */
-    explicit sigh_mixin(const allocator_type &allocator)
-        : Type{allocator},
+    explicit sigh_mixin(const allocator_type &allocator, traits_type traits = {})
+        : Type{allocator, std::move(traits)},
           owner{},
           construction{allocator},
           destruction{allocator},
--- a/src/entt/entity/registry.hpp
+++ b/src/entt/entity/registry.hpp
@@ -247,23 +247,41 @@ class basic_registry {
     using pool_container_type = dense_map<id_type, std::shared_ptr<base_type>, identity, std::equal_to<id_type>, typename alloc_traits::template rebind_alloc<std::pair<const id_type, std::shared_ptr<base_type>>>>;
     using group_container_type = dense_map<id_type, std::shared_ptr<internal::group_descriptor>, identity, std::equal_to<id_type>, typename alloc_traits::template rebind_alloc<std::pair<const id_type, std::shared_ptr<internal::group_descriptor>>>>;
 
+    template<typename Type, typename AllocatorType>
+    void create_storage(const id_type id, const AllocatorType &allocator,
+                        component_traits<Type> traits) {
+        static_assert(std::is_same_v<Type, std::decay_t<Type>>, "Non-decayed types not allowed");
+
+        auto &cpool = pools[id];
+        ENTT_ASSERT(!cpool, "Expected the pool not to be set!");
+
+        cpool = std::allocate_shared<storage_for_type<std::remove_const_t<Type>>>(get_allocator(), allocator, std::move(traits));
+        cpool->bind(forward_as_any(*this));
+
+        ENTT_ASSERT(!!cpool, "Expected the pool to be set after construction!");
+    }
+
     template<typename Type>
     [[nodiscard]] auto &assure(const id_type id = type_hash<Type>::value()) {
         static_assert(std::is_same_v<Type, std::decay_t<Type>>, "Non-decayed types not allowed");
+
         auto &cpool = pools[id];
 
         if(!cpool) {
             using storage_type = storage_for_type<Type>;
-            using alloc_type = typename storage_type::allocator_type;
+            using storage_allocator = typename storage_type::allocator_type;
+            using storage_traits_type = typename storage_type::traits_type;
 
-            if constexpr(std::is_same_v<Type, void> && !std::is_constructible_v<alloc_type, allocator_type>) {
+            if constexpr(!std::is_constructible_v<storage_allocator, allocator_type const &>) {
                 // std::allocator<void> has no cross constructors (waiting for C++20)
-                cpool = std::allocate_shared<storage_type>(get_allocator(), alloc_type{});
+                // Afterwards, the ability to default construct the storage_allocator that
+                // are not compatible to the owning registries allocator_type must be preserved.
+                create_storage<Type>(id, storage_allocator{}, storage_traits_type{});
             } else {
-                cpool = std::allocate_shared<storage_type>(get_allocator(), get_allocator());
+                create_storage<Type>(id, allocator_type{get_allocator()}, storage_traits_type{});
             }
 
-            cpool->bind(forward_as_any(*this));
+            ENTT_ASSERT(!!cpool, "Expected the pool to be set after construction!");
         }
 
         ENTT_ASSERT(cpool->type() == type_id<Type>(), "Unexpected type");
@@ -444,6 +462,35 @@ public:
     template<typename Type>
     const storage_for_type<Type> *storage(const id_type id = type_hash<Type>::value()) const {
         return assure<Type>(id);
+    }
+
+    /**
+     * @brief Tries to create a storage for a component with the given arguments.
+     *
+     * @warning
+     * If a storage for the given component exist already, false is returned!
+     *
+     * @tparam Type Type of component of which to return the storage.
+     * @param allocator The allocator passed to the storage.
+     * @param traits Optional component traits type passed to the storage.
+     * @param id Optional name used to map the storage within the registry.
+     * @return Returns true if the component could be created successfully.
+     *         If the component existed already the creation is skipped and false is returned.
+     */
+    template<typename Type>
+    bool try_create_storage(const typename storage_for_type<Type>::allocator_type &allocator,
+                            component_traits<Type> traits = {},
+                            const id_type id = type_hash<Type>::value()) {
+        static_assert(std::is_same_v<Type, std::decay_t<Type>>, "Non-decayed types not allowed");
+
+        if(!!storage(id)) {
+            // If the storage already exists return false
+            return false;
+        }
+
+        // Otherwise create the storage
+        create_storage(id, allocator, traits);
+        return true;
     }
 
     /**
--- a/src/entt/entity/storage.hpp
+++ b/src/entt/entity/storage.hpp
@@ -341,7 +341,7 @@ class basic_storage
 
         if(!(idx < payload.size())) {
             auto curr = payload.size();
-            allocator_type allocator{get_allocator()};
+            allocator_type &allocator = get_allocator();
             payload.resize(idx + 1u, nullptr);
 
             ENTT_TRY {
@@ -371,7 +371,7 @@ class basic_storage
 
         ENTT_TRY {
             auto elem = assure_at_least(static_cast<size_type>(it.index()));
-            allocator_type allocator{get_allocator()};
+            allocator_type &allocator = get_allocator();
             entt::uninitialized_construct_using_allocator(to_address(elem), allocator, std::forward<Args>(args)...);
         }
         ENTT_CATCH {
@@ -384,7 +384,7 @@ class basic_storage
 
     void shrink_to_size(const std::size_t sz) {
         const auto from = (sz + traits_type::page_size - 1u) / traits_type::page_size;
-        allocator_type allocator{get_allocator()};
+        allocator_type &allocator = get_allocator();
 
         for(auto pos = sz, length = base_type::size(); pos < length; ++pos) {
             if constexpr(traits_type::in_place_delete) {
@@ -417,7 +417,7 @@ private:
 
             if constexpr(traits_type::in_place_delete) {
                 if(base_type::operator[](to) == tombstone) {
-                    allocator_type allocator{get_allocator()};
+                    allocator_type &allocator = get_allocator();
                     entt::uninitialized_construct_using_allocator(to_address(assure_at_least(to)), allocator, std::move(elem));
                     alloc_traits::destroy(allocator, std::addressof(elem));
                     return;
@@ -440,7 +440,7 @@ protected:
      * @param last An iterator past the last element of the range of entities.
      */
     void pop(underlying_iterator first, underlying_iterator last) override {
-        for(allocator_type allocator{get_allocator()}; first != last; ++first) {
+        for(allocator_type &allocator = get_allocator(); first != last; ++first) {
             // cannot use first.index() because it would break with cross iterators
             auto &elem = element_at(base_type::index(*first));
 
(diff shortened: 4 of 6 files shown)


Naios:
@skypjack From my side the initial step on this PR is finished. I would add any missing unit tests after receiving your feedback.

Currently there are two main points that would require a review:

- The workaround for polymorphic allocators which is needed because `std::polymorphic_allocator::construct` (at least on MSVC) does not propagate this correctly prior to C++20.
- I noticed that `basic_storage::pop` calls `std::exchange` and therefore `traits::move_element` on the same both elements (the objects can alias). While this is theoretically valid it would be better if those aliasing calls are not made by the algorithm in the first place. There is also a very small chance that this behaviour is caused by my changes.

I'm looking forward to your feedback.

Naios:
@skypjack Do you have any news about this PR? I have updated it to the latest `wip` branch commit and improved the previously mentioned allocator copying.

skypjack:
> @skypjack Do you have any news about this PR? Would you prefer any possible changes to it? I have updated it to the latest `wip` branch commit and improved the previously mentioned allocator copying.

Oh, sorry. I saw the todo list with some unchecked bullets yet and thought you might want to continue before reviewing. I'll look into it as soon as possible then and drop my thoughts here.
I already had a look after the first commits to be honest and I must admit that very invasive changes to key library classes (like the storage) to support minor feature don't drive me crazy.
I expected something less... impactful, maybe? You know, `EnTT` is entirely designed to offer hook points and build up _from the outside_. Just look at the signal support for example, which in itself is an external class and not embedded in storage.
Just out of curiosity, have you tried this approach already? If yes, what problems stopped you and made you turn towards a built-in solution for the storage?

src/entt/entity/storage.hpp
```diff
@@ -489,8 +610,24 @@ class basic_storage: public basic_sparse_set<Entity, typename std::allocator_tra
      * @brief Returns the associated allocator.
      * @return The associated allocator.
      */
-    [[nodiscard]] constexpr allocator_type get_allocator() const noexcept {
-        return payload.get_allocator();
+    [[nodiscard]] constexpr allocator_type &get_allocator() noexcept {
```
skypjack, line 613: This is a huge no. There is no reason to return a reference to an internal allocator which isn't even required to exist.
`get_allocator` is defined as `allocator_type get_allocator() const noexcept` literally everywhere and for a good reason.

src/entt/entity/storage.hpp
```diff
@@ -110,45 +118,101 @@ class storage_iterator final {
...
+    }
+
+    Container *container{};
+};
+
+/**
+ * @brief Storage accessor for dynamically sized components
+ */
+template<typename Storage>
+struct dynamic_storage_accessor {
```
skypjack, line 192: This is yet another level of indirection and an attempt to hide some details a little further. However, there exists already the traits class.
You can provide the iterator with the container (as it was before), then use the traits to get the job done if you want to hide some details away from the iterator itself.
More classes with a very similar purpose means more things to maintain that are related but also spread around for no good reasons.

src/entt/entity/storage.hpp
```diff
@@ -210,6 +274,16 @@ template<typename... Lhs, typename... Rhs>
     return !(lhs == rhs);
 }
 
+template<typename Traits, typename = void>
+struct is_component_trait_dynamically_sized
+    : std::false_type {};
+
+template<typename Traits>
```
skypjack, line 281: These are used only once. Here:
```cpp
static constexpr bool is_dynamically_sized_type_v = internal::is_component_trait_dynamically_sized<component_traits<Type>>::value;
```
Where the following would be enough anyway:
```cpp
static constexpr bool is_dynamically_sized_type_v = component_traits<Type>::has_dynamic_size;
```
To me, this class seems completely useless. Just yet another thing to maintain for no good reasons. Isn't it?

src/entt/entity/storage.hpp
```diff
@@ -233,25 +307,41 @@ template<typename... Lhs, typename... Rhs>
  * @tparam Allocator Type of allocator used to manage memory and elements.
  */
 template<typename Type, typename Entity, typename Allocator, typename>
-class basic_storage: public basic_sparse_set<Entity, typename std::allocator_traits<Allocator>::template rebind_alloc<Entity>> {
+class basic_storage
+    : public basic_sparse_set<Entity, typename std::allocator_traits<Allocator>::template rebind_alloc<Entity>>,
+      protected Allocator,
```
skypjack, line 312: Inheriting from an allocator and a traits class sounds soooooo bad.
Why aren't they just members? I don't see any benefit from inheritance. here.

src/entt/entity/storage.hpp
```diff
@@ -233,25 +307,41 @@ template<typename... Lhs, typename... Rhs>
...
+      protected component_traits<Type> {
+    template<typename Container, std::size_t Size>
+    friend struct internal::static_storage_accessor;
+    template<typename Storage>
+    friend struct internal::dynamic_storage_accessor;
```
skypjack, line 317: These `friend`s are a code smell that makes me wonder again if these _accessors_ aren't superfluous.

src/entt/entity/storage.hpp
```diff
@@ -233,25 +307,41 @@ template<typename... Lhs, typename... Rhs>
...
+        }
     }
 
     auto assure_at_least(const std::size_t pos) {
         const auto idx = pos / traits_type::page_size;
 
         if(!(idx < payload.size())) {
             auto curr = payload.size();
-            allocator_type allocator{get_allocator()};
+            allocator_type &allocator = get_allocator();
```
skypjack, line 344: No way you should have a `get_allocator` function returning a reference. Really.

src/entt/entity/storage.hpp
```diff
@@ -425,16 +535,21 @@ class basic_storage: public basic_sparse_set<Entity, typename std::allocator_tra
     /*! @brief Constant extended iterable storage proxy. */
     using const_iterable = iterable_adaptor<internal::extended_storage_iterator<typename base_type::const_iterator, const_iterator>>;
 
+    using underlying_type::type;
```
skypjack, line 538: ??

src/entt/entity/storage.hpp
```diff
@@ -793,16 +932,20 @@ class basic_storage<Type, Entity, Allocator, std::enable_if_t<component_traits<T
     /*! @brief Constant extended iterable storage proxy. */
     using const_iterable = iterable_adaptor<internal::extended_storage_iterator<typename base_type::const_iterator>>;
 
+    using underlying_type::type;
```
skypjack, line 935: ??

src/entt/entity/storage.hpp
```diff
@@ -833,6 +977,14 @@ class basic_storage<Type, Entity, Allocator, std::enable_if_t<component_traits<T
         return allocator_type{base_type::get_allocator()};
     }
 
+    /**
+     * @brief Returns the associated traits.
+     * @return The associated traits.
+     */
+    [[nodiscard]] constexpr traits_type const &get_traits() const noexcept {
```
skypjack, line 984: Does it make sense to have traits on an empty type? It sounds like an unwanted side effect. 🤔

src/entt/entity/storage.hpp
```diff
@@ -1035,6 +1192,14 @@ class basic_storage<Entity, Entity, Allocator>
         return *this;
     }
 
+    /**
+     * @brief Returns the associated traits.
+     * @return The associated traits.
+     */
+    [[nodiscard]] constexpr traits_type const &get_traits() const noexcept {
```
skypjack, line 1199: Same as above, what's the purpose of having traits on this specialization?

skypjack (commented):
I've had a chance to take a look at it and I think there are several things that need to be sorted out before it gets anywhere near the review.
However, I want to understand one thing first. From what I see from the tests, the whole thing looks no different than this:

* I create a storage with a _fake_ type which basically helps to make my storage understand that I have to pass it a size and that it has to do some things differently than usual, as well as use a certain type as the _base type_ for buffers of the requested size (`uint8_t` in your case but I don't even know how much that makes sense within this context)

* The storage, at the end of the day, works with buffer objects of the size I requested and returns them to me on request, not caring what I really do with them afterwards (I can build an object of type `T` or `U` within them or just use them as arrays of plain integral values, it doesn't change anything)

* Building views or anything else with these objects must be possible (and I'd be amazed otherwise)

* When I tell the storage to destroy an object, just pray that I call `~U()` on it first to avoid problems because the storage (rightfully) doesn't know anything about my `U`

Did I miss something? Ignoring the implementation details, is this the end goal?
Can you elaborate a little further on it otherwise? Because a proper description of what we're trying to achieve here is missing.

skypjack:
> However, I made the changes as they are currently proposed because I think that supporting dynamic component sizes is probably a first level feature you would like to support.

For sure. Like supporting empty types or entity storage with special abilities. Yet, in both cases, this was through specialziations.
So I was wondering why dynamic type support can't go the same way. I haven't stopped to think about it for long, so I might not see all the problems. However, something like `storage<dynamic<32>>` seems doable.
What drawbacks does it have? Maybe you've already thought about it or tried it and you know it better than me.

Naios:
@skypjack Thank you for your comprehensive feedback.

---

> When I tell the storage to destroy an object, just pray that I call ~U() on it first to avoid problems because the storage (rightfully) doesn't know anything about my U

I think this a misconception: The allocator is responsible for construction and destruction and can be adapted to the stored object. Therefore it is entirely responsible for the objects lifetime and the objects constructor and destructor are always called. This is simply not the case in the test code which is currently only specialized for primitive destructible objects.

By moving the responsibility to the allocator, various object types can be supported. For instance also dynamically generated and reflected objects.

---

> For sure. Like supporting empty types or entity storage with special abilities. Yet, in both cases, this was through specialziations.
So I was wondering why dynamic type support can't go the same way.

There are no technical limitations I'm aware of.
I chose to modify the existing storage to reduce code duplication.

The big issue I see is that the storage is mostly a facade to the underlying container, therefore it probably makes sence to specialize the container rather than the storage. I will look into this.
The reason why the trait is made staeful in this PR is that we need an object that can be specialized for the component type. And the trait itself (if we ignore its naming) fits the role perfectly because it effectively specialized the storage based on the component type already (page size, empty type). Why not use it for any other use cases where the storage needs to be specialized?

src/entt/core/memory.hpp
```diff
@@ -222,6 +222,16 @@ struct uses_allocator_construction<std::pair<Type, Other>> {
...
 };
 
+/**
+ * @brief deduces to a true type if the allocator provides a resource method.
+ */
+template<typename Allocator, typename = void>
+struct is_polymorphic_allocator: std::false_type {
+};
+template<typename Allocator>
+struct is_polymorphic_allocator<Allocator, std::void_t<decltype(std::declval<Allocator &>().resource())>>: std::true_type {
```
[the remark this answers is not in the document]
Naios: This can be done, but I did not like to include the `<memory_resource>` header. I can change this based on your preference.

src/entt/core/memory.hpp
```diff
@@ -280,8 +290,25 @@ constexpr Type make_obj_using_allocator(const Allocator &allocator, Args &&...ar
...
  */
 template<typename Type, typename Allocator, typename... Args>
-constexpr Type *uninitialized_construct_using_allocator(Type *value, const Allocator &allocator, Args &&...args) {
-    return std::apply([value](auto &&...curr) { return new(value) Type(std::forward<decltype(curr)>(curr)...); }, internal::uses_allocator_construction<Type>::args(allocator, std::forward<Args>(args)...));
+constexpr Type *uninitialized_construct_using_allocator(Type *value, Allocator &allocator, Args &&...args) {
+    return std::apply([value, &allocator](auto &&...curr) mutable -> Type * {
+        using alloc_type = std::decay_t<Allocator>;
+        using alloc_traits = std::allocator_traits<alloc_type>;
+
+        static_assert(std::is_convertible_v<Type *, typename alloc_traits::value_type *>,
```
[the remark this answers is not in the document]
Naios: No its not redundant, it verifies the type definitions of the specialized allocator.

src/entt/core/memory.hpp
```diff
@@ -280,8 +290,25 @@ constexpr Type make_obj_using_allocator(const Allocator &allocator, Args &&...ar
  * @return A pointer to the newly created object of the given type.
  */
 template<typename Type, typename Allocator, typename... Args>
-constexpr Type *uninitialized_construct_using_allocator(Type *value, const Allocator &allocator, Args &&...args) {
-    return std::apply([value](auto &&...curr) { return new(value) Type(std::forward<decltype(curr)>(curr)...); }, internal::uses_allocator_construction<Type>::args(allocator, std::forward<Args>(args)...));
+constexpr Type *uninitialized_construct_using_allocator(Type *value, Allocator &allocator, Args &&...args) {
```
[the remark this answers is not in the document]
Naios: This change should be semantically equal for the default allocator but it makes use of the `construct_at` provided by a specialized allocator.

src/entt/entity/mixin.hpp
```diff
@@ -89,9 +91,10 @@ class sigh_mixin final: public Type {
     /**
      * @brief Constructs an empty storage with a given allocator.
      * @param allocator The allocator to use.
+     * @param traits The type_traits to use.
      */
-    explicit sigh_mixin(const allocator_type &allocator)
-        : Type{allocator},
+    explicit sigh_mixin(const allocator_type &allocator, traits_type traits = {})
```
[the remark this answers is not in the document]
Naios: The reason why the trait is made staeful in this PR is that we need an object that can be specialized for the component type. And the trait itself (if we ignore its naming) fits the role perfectly because it effectively specialized the storage based on the component type already (page size, empty type). Why not use it for any other use cases where the storage needs to be specialized?

src/entt/entity/registry.hpp
```diff
@@ -247,23 +247,41 @@ class basic_registry {
...
+                        component_traits<Type> traits) {
+        static_assert(std::is_same_v<Type, std::decay_t<Type>>, "Non-decayed types not allowed");
+
+        auto &cpool = pools[id];
+        ENTT_ASSERT(!cpool, "Expected the pool not to be set!");
+
+        cpool = std::allocate_shared<storage_for_type<std::remove_const_t<Type>>>(get_allocator(), allocator, std::move(traits));
+        cpool->bind(forward_as_any(*this));
+
+        ENTT_ASSERT(!!cpool, "Expected the pool to be set after construction!");
```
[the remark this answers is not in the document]
Naios: `std::allocate_shared` still can call into a specialized allocator as far as I'm aware of.

src/entt/entity/registry.hpp
```diff
@@ -446,6 +464,35 @@ class basic_registry {
...
+     * If a storage for the given component exist already, false is returned!
+     *
+     * @tparam Type Type of component of which to return the storage.
+     * @param allocator The allocator passed to the storage.
+     * @param traits Optional component traits type passed to the storage.
+     * @param id Optional name used to map the storage within the registry.
+     * @return Returns true if the component could be created successfully.
+     *         If the component existed already the creation is skipped and false is returned.
+     */
+    template<typename Type>
```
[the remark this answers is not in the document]
Naios: The reason why this function cannot be factored into the stroage is that it would be very useful to  check whether the storage was freshly created with exact the provided arguments to verify correct initialization.

Since `storage` always returns the (maybe already default constructed storage) the `storage` method is not sufficient.

The default initializatin always needs to be possible for any storage by library design.

src/entt/entity/storage.hpp
Naios: I agree, this needs to be changed.

src/entt/entity/storage.hpp
Naios: `get_allocator` can be changed back to a return-by-value, however the internal use of `get_allocator` needs to be changed to a retrieve by reference because the allocator can be stateful and in this case a repretive call to `get_allocator` could add overhead.

src/entt/entity/storage.hpp
Naios: I try to remove this specialization by moving everything into a specialized container.

src/entt/entity/storage.hpp
Naios: This is to enable empty type optimization, I can changed it as mentioned in one of my initial comments above.

src/entt/entity/storage.hpp
Naios: See above retreieving the allocator by-value can add overhead for stateful allocators.

src/entt/entity/storage.hpp
Naios: Required for deduplication.

src/entt/entity/storage.hpp
Naios: This method exists for consistency with the other specializations.

src/entt/core/memory.hpp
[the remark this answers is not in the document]
skypjack: Yeah, I mean, it would break later on in any case if this condition wasn't satisfied. Isn't it the case?

src/entt/core/memory.hpp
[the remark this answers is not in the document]
skypjack: There is a compile-time  branch now. How can it be semantically equal? I'm pretty sure one can easily pack an example that works in one case and behaves differently in the other one. The whole change would be pointless otherwise. 🤔

src/entt/entity/mixin.hpp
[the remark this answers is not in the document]
skypjack: I think _because it leaks to classes that don't care about it at all_ is my first objection to _using it for any other use cases where the storage need to be specialized_.
In terms of software design, I generally look suspiciously at my code if it leaks private information from a base class to a derived class. This is exactly the case from what I get. For example, the `sigh_mixin` as well as any other mixin doesn't care about these traits.
Also, does it mean that all existing mixin out there must be updated accordingly? Because I've a bunch of them in production code that I can't easily touch/change/update.

src/entt/entity/registry.hpp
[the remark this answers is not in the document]
skypjack: Well, in this case you want to specialize the allocator though. Why stressing `allocate_shared`? 🤔 
I'm missing something probably but it still seems superfluous. Isn't it?

src/entt/entity/registry.hpp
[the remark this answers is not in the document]
skypjack: I might be wrong but this looks like a direct consequence of the choice of using runtime traits as a constructor argument.
If the runtime values were on the storage itself, you could create the latter and set them without the need of an extra function.

src/entt/entity/storage.hpp
skypjack: The API and the internal implementation are different beasts. I pocked more than once in the standard library implementation and I'm always amazed by the amount of... code 😅 that lies behind an API sometimes.

src/entt/entity/storage.hpp
skypjack: I specifically implemented and used the `compressed_pair` class for the same purpose and the main goal was to exploit ETO without having to inherit from an allocator. 👍🙂

src/entt/entity/storage.hpp
skypjack: Can you elaborate a little further? I think I'm missing some context here.

skypjack:
Closed as starving/no longer viable.

PR #1011, opened 2023-05-06
``std::tuple`` trait specializations for ``entt::type_list`` and ``entt::value_list``

DNKpp:
As discussed on the discord, I would like to specialize these common traits for the entt types.

As a second minor improvement, I unified your specializations of ``type_list_transform`` into one specialization. If you don't like it, I'll revert that change.

Hopefully these changes match your style-guide. Resharper doesn't seem to recognize the clang-format correctly.

### DIFF
--- a/src/entt/core/tuple.hpp
+++ b/src/entt/core/tuple.hpp
@@ -110,4 +110,7 @@ struct std::tuple_element<Index, entt::type_list<Type...>>: entt::type_list_elem
 template <auto... Value>
 struct std::tuple_size<entt::value_list<Value...>>: std::integral_constant<std::size_t, entt::value_list<Value...>::size> {};
 
+template <std::size_t Index, auto... Value>
+struct std::tuple_element<Index, entt::value_list<Value...>>: entt::value_list_element<Index, entt::value_list<Value...>> {};
+
 #endif
--- a/src/entt/core/tuple.hpp
+++ b/src/entt/core/tuple.hpp
@@ -107,4 +107,7 @@ struct std::tuple_size<entt::type_list<Type...>>: std::integral_constant<std::si
 template <std::size_t Index, typename... Type>
 struct std::tuple_element<Index, entt::type_list<Type...>>: entt::type_list_element<Index, entt::type_list<Type...>> {};
 
+template <auto... Value>
+struct std::tuple_size<entt::value_list<Value...>>: std::integral_constant<std::size_t, entt::value_list<Value...>::size> {};
+
 #endif
--- a/src/entt/core/tuple.hpp
+++ b/src/entt/core/tuple.hpp
@@ -4,6 +4,7 @@
 #include <tuple>
 #include <type_traits>
 #include <utility>
+#include "type_traits.hpp"
 
 namespace entt {
 
@@ -99,5 +100,8 @@ template<typename Func>
 forward_apply(Func) -> forward_apply<std::remove_reference_t<std::remove_cv_t<Func>>>;
 
 } // namespace entt
+
+template <typename... Type>
+struct std::tuple_size<entt::type_list<Type...>>: std::integral_constant<std::size_t,  entt::type_list<Type...>::size> {};
 
 #endif
--- a/src/entt/core/tuple.hpp
+++ b/src/entt/core/tuple.hpp
@@ -102,6 +102,9 @@ forward_apply(Func) -> forward_apply<std::remove_reference_t<std::remove_cv_t<Fu
 } // namespace entt
 
 template <typename... Type>
-struct std::tuple_size<entt::type_list<Type...>>: std::integral_constant<std::size_t,  entt::type_list<Type...>::size> {};
+struct std::tuple_size<entt::type_list<Type...>>: std::integral_constant<std::size_t, entt::type_list<Type...>::size> {};
+
+template <std::size_t Index, typename... Type>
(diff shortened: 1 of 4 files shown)


src/entt/core/tuple.hpp
```diff
@@ -100,4 +101,16 @@ forward_apply(Func) -> forward_apply<std::remove_reference_t<std::remove_cv_t<Fu
 
 } // namespace entt
 
+template <typename... Type>
+struct std::tuple_size<entt::type_list<Type...>>: std::integral_constant<std::size_t, entt::type_list<Type...>::size> {};
```
skypjack, line 105: Not sure this is the right place for these. 🤔 
`type_traits.hpp` seems more appropriate actually.

src/entt/core/type_traits.hpp
```diff
@@ -335,13 +335,14 @@ struct type_list_transform;
 
 /**
  * @brief Applies a given _function_ to a type list and generate a new list.
+ * @tparam Container Template type containing types.
  * @tparam Type Types provided by the type list.
  * @tparam Op Unary operation as template class with a type member named `type`.
  */
-template<typename... Type, template<typename...> class Op>
-struct type_list_transform<type_list<Type...>, Op> {
+template<template<typename...> class Container, typename... Type, template<typename...> class Op>
```
skypjack, line 342: I see the point of this one. However, as an user, I would be surprised to see the following to working properly:
```cpp
using type = type_list_transform_t<std::tuple<T, U, V>, my_op>;
```
I'd be even more surprised by the sheer amount of errors and subtleties it can throw at me.

skypjack (changes requested):
To be honest, I'm fine with the `tuple_size` and `tuple_element` specialziations for `type_list` and `value_list`.
On the other hand, I'm a bit uncomfortable with the more _flexible_ version of `type_list_transform` because I see how it can give lot of headaches to users.
I suggest to limit the changes to the former, unless there is a good reason to also include the latter.

src/entt/core/type_traits.hpp
DNKpp: Well, how about adding then ``static_assert(std::is_base_of_v<type_list<Type...>, Container<Type...>>, "The provided List must be entt::type_list or a derived class.")`` or making the ``type`` alias SFINAE
``using type = std::enable_if_t<std::is_base_of_v<type_list<Type...>, Container<Type...>>, Container<typename Op<Type>::type...>>``

src/entt/core/tuple.hpp
DNKpp: Well, that was your suggestion back then ^^

DNKpp:
> [...] I'm a bit uncomfortable with the more _flexible_ version of `type_list_transform` because I see how it can give lot of headaches to users. I suggest to limit the changes to the former, unless there is a good reason to also include the latter.

Currently I kind of dislike the approach to specialize all the ``type_list``-algorithms for any derived class of ``entt::type_list``, because that quickly gets out of hands. Expecially as you explicitly allow and expect ``entt::type_list`` to be derived from. As a user I would be then very surprised to not be able to query
```cpp
entt::type_list_unique_t<entt::excluded_t<...>>
```
This seems to me rather a workaround than a real feature.

src/entt/core/tuple.hpp
skypjack: Yeah, I know 😅 I changed my mind though, it wasn't a good suggestion actually.

skypjack:
> Currently I kind of dislike the approach to specialize all the `type_list`-algorithms for any derived class of `entt::type_list`, because that quickly gets out of hands.

That's how the standard library works though, which is something I try to stick with in general.
Look at `tuple_element` for example. It's not defined, then specializations for the internal relevant classes are provided.
I like this approach more to be honest. I prefer adding a couple of specializations rather than trying to explain over and over on GH that the user code fails because someone tried to use the tool with the _wrong_ type.
You're an advanced C++ user, I'm pretty sure you can figure it out on your own and fix the problem. However, lot of times someone comes and says - _ehy, the error is in the library codebase, thus it's a bug of the library_.
It's really hard to explain that this isn't the case. On the other side, if the stack trace drives you in your own codebase... 🤷‍♂️😅

DNKpp:
> That's how the standard library works though, which is something I try to stick with in general.
> Look at tuple_element for example. It's not defined, then specializations for the internal relevant classes are provided.

Yeah, that's right. But on the other hand-side the standard doesn't expect someone to inherit from ``std::tuple`` or something like that. ``entt`` does expect and even does it itself.

> I prefer adding a couple of specializations rather than trying to explain over and over on GH that the user code fails because someone tried to use the tool with the wrong type.

I fully understand your point. That's why I came up with the checking approach (via ``static_assert`` or ``SFINAE`` the ``type`` alias). Did you see that? https://github.com/skypjack/entt/pull/1011#discussion_r1187214679

Either way. If you dislike that I'm fine and will revert that. Just trying leech some of your experience :)

skypjack:
Oh, it's not a matter of experience 😅 it's more about my mental illness.
And yeah, I prefer to stick with the same approach used so far. I see your point and completely understand it though, do not get me wrong. It's just that 🤷‍♂️ dunno, I find the std-like behavior more _natural_, if you get what I mean.
Thank you and sorry for the late reply. 🙏

DNKpp:
Well, I reverted the change. It should be finished then :)

skypjack (approved):

PR #1014, opened 2023-05-10
Fix organizer::vertex::prepare not creating component pools

### LINKED ISSUE #1013: entt::organizer::vertex::prepare does not create the required component pools
The documentation for entt::organizer mentions this:

> Since the creation of pools and resources within the registry isn't necessarily thread safe, each vertex also offers a `prepare` function which is used to setup a registry for execution with the created graph

However, the current implementation of `prepare` only successfully creates missing context variables, and does not create the component pools required by any "view" parameters of the user-provided systems. Here is a minimal example that demonstrates the bug:
```cpp
#include <entt/entt.hpp>
#include <iostream>

struct test_component {};

void test_system(entt::view<entt::get_t<test_component>> view) {
    (void)view;
}

int main() {
    entt::organizer organizer{};
    organizer.emplace<test_system>();

    entt::registry registry{};
    for (const entt::organizer::vertex& node : organizer.graph()) {
        node.prepare(registry);
    }

    const auto& s = registry.storage();
    auto it = s.begin();
    const auto end = s.end();
    ++it; // Skip "shortcut" storage.
    if (it != end) {
        std::cout << "good\n";
    } else {
        std::cout << "bad\n";
    }
}
```
This code currently prints "bad", which indicates that the registry contains no component pools except for the initial "shortcut" pool of type `entity_type`, despite `test_system` requiring `test_component`. If the `prepare` loop is replaced with simply `(void)registry.view<test_component>();`, then the program prints "good" as expected.

After digging into the code, I believe the main cause of the bug is in the `extract` function on line 129 of `organizer.hpp`: https://github.com/skypjack/entt/blob/master/src/entt/entity/organizer.hpp#L129
The intent here seems to be that the temporary `as_view` object should be implicitly converted to the concrete view type in order to invoke its internal `dispatch` method which is what actually constructs the view (and, consequently, its required pools): https://github.com/skypjack/entt/blob/master/src/entt/entity/helper.hpp#L48
Unfortunately, this conversion only happens when the actual callback function is called using `std::apply` with the result of `to_args` at https://github.com/skypjack/entt/blob/master/src/entt/entity/organizer.hpp#L300, and not when the result of `to_args` is discarded, which is what `prepare` does: https://github.com/skypjack/entt/blob/master/src/entt/entity/organizer.hpp#L310

A possible fix could be to explicitly construct the concrete view type inside of `extract` like this:
```cpp
return Type{as_view{reg}};
```

DonutVikingChap:
Fixes #1013

### DIFF
--- a/src/entt/entity/organizer.hpp
+++ b/src/entt/entity/organizer.hpp
@@ -126,7 +126,7 @@ class basic_organizer final {
         if constexpr(std::is_same_v<Type, Registry>) {
             return reg;
         } else if constexpr(internal::is_view_v<Type>) {
-            return as_view{reg};
+            return static_cast<Type>(as_view{reg});
         } else {
             return reg.ctx().template emplace<std::remove_reference_t<Type>>();
         }
--- a/test/entt/entity/organizer.cpp
+++ b/test/entt/entity/organizer.cpp
@@ -1,5 +1,4 @@
 #include <cstddef>
-#include <iterator>
 #include <gtest/gtest.h>
 #include <entt/core/type_info.hpp>
 #include <entt/entity/organizer.hpp>
--- a/test/entt/entity/organizer.cpp
+++ b/test/entt/entity/organizer.cpp
@@ -1,4 +1,5 @@
 #include <cstddef>
+#include <utility>
 #include <gtest/gtest.h>
 #include <entt/core/type_info.hpp>
 #include <entt/entity/organizer.hpp>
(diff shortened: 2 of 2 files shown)


skypjack:
Oh, I'm fine with this PR but the changes to the test are waaaaay easier 🙂 this should be enough actually:
```cpp
TEST(Organizer, Prepare) {
    entt::organizer organizer;
    entt::registry registry;
    clazz instance;

    organizer.emplace<&ro_int_double>();
    organizer.emplace<&clazz::rw_int_char>(instance);

    const auto graph = organizer.graph();

    ASSERT_EQ(std::as_const(registry).storage<int>(), nullptr);
    ASSERT_EQ(std::as_const(registry).storage<double>(), nullptr);
    ASSERT_EQ(std::as_const(registry).storage<char>(), nullptr);

    ASSERT_FALSE(registry.ctx().contains<int>());
    ASSERT_FALSE(registry.ctx().contains<char>());

    for(auto &&vertex: graph) {
        vertex.prepare(registry);
    }

    ASSERT_NE(std::as_const(registry).storage<int>(), nullptr);
    ASSERT_NE(std::as_const(registry).storage<char>(), nullptr);

    ASSERT_FALSE(registry.ctx().contains<int>());
    ASSERT_FALSE(registry.ctx().contains<char>());
    ASSERT_TRUE(registry.ctx().contains<double>());
}
```
Checking that the storage exists after invoking `prepare` only takes a couple of lines. 👍

DonutVikingChap:
That would be a lot cleaner, but storage<T>() doesn't return a pointer, it just returns a static placeholder pool when the requested pool doesn't exist, and since both the placeholder and the actual pool should be empty, there's no easy way to check that result. Iterating over the existing pools was the only way I could think of that works around that, but maybe the registry API could be extended with a separate function that checks whether a pool has been created or not?

skypjack:
> it just returns a static placeholder pool when the requested pool doesn't exist

Oh, that's because the PR points to `master`. Wait. I'm switching it to `wip` which is the upcoming `master`.

DonutVikingChap:
Ok! The test is updated now.

PR #1034, opened 2023-07-05
dense_*: workaround for the max() ambiguity with MAX macro.

pavelpokutnev:
There is ambiguity between std::max() and MAX macro, which propagates via "Windows.h", on the Windows platform. In order not to bother the user with the NOMINMAX definitions, the bracketing method was chosen.


pavelpokutnev:
OK, just update the PR.

PR #1047, opened 2023-07-31
Fix C++20 global module fragment inclusion

DonutVikingChap:
Using the experimental C++20 modules support in recent builds of Clang, EnTT currently causes compilation errors when included as part of the global module fragment in a module interface unit that exports part of EnTT's API and is then imported and used by a different translation unit. Here is a minimal example (note the "error: no matching function for call to 'popcount'" in the output): https://godbolt.org/z/9vTjz9Prh

The reason seems to be that certain function declarations that are implicitly required by the implementation aren't exported into the module interface since they are erroneously marked 'static' and have internal linkage when they should (presumably) be 'inline'. I found that changing these declarations accordingly fixed the problem and made EnTT work well with C++20 modules in my project, and this PR reflects those changes.

Unfortunately I can't write proper test cases for this since it depends on features that are still experimental in most compilers, so there may still be other static declarations that I have missed that could cause problems in other modules-based projects, but as long as these changes don't break anything for regular non-modules builds, they should hopefully at least help make EnTT a bit more modules-friendly in the short term.

### DIFF
--- a/src/entt/core/type_info.hpp
+++ b/src/entt/core/type_info.hpp
@@ -44,7 +44,7 @@ template<typename Type, auto = stripped_type_name<Type>().find_first_of('.')>
 }
 
 template<typename Type>
-[[nodiscard]] inline std::string_view type_name(char) noexcept {
+[[nodiscard]] std::string_view type_name(char) noexcept {
     static const auto value = stripped_type_name<Type>();
     return value;
 }
@@ -57,7 +57,7 @@ template<typename Type, auto = stripped_type_name<Type>().find_first_of('.')>
 }
 
 template<typename Type>
-[[nodiscard]] inline id_type type_hash(char) noexcept {
+[[nodiscard]] id_type type_hash(char) noexcept {
     static const auto value = [](const auto stripped) {
         return hashed_string::value(stripped.data(), stripped.size());
     }(stripped_type_name<Type>());
--- a/src/entt/core/type_info.hpp
+++ b/src/entt/core/type_info.hpp
@@ -38,26 +38,26 @@ template<typename Type>
 }
 
 template<typename Type, auto = stripped_type_name<Type>().find_first_of('.')>
-[[nodiscard]] static constexpr std::string_view type_name(int) noexcept {
+[[nodiscard]] constexpr std::string_view type_name(int) noexcept {
     constexpr auto value = stripped_type_name<Type>();
     return value;
 }
 
 template<typename Type>
-[[nodiscard]] static std::string_view type_name(char) noexcept {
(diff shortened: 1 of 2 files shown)


skypjack:
Makes sense actually, thanks for pointing this out. Why `inline` though? We can just drop the `static` and that's it.
Those are template functions, no need to make them inline to avoid symbols duplication. On the other hand, I'm pretty sure that the compiler knows better than me whether to inline them or not.
So, yeah, I'd rather get rid of it if possible before merging. Thanks. 👍

src/entt/core/type_info.hpp
```diff
@@ -38,26 +38,26 @@ template<typename Type>
...
 template<typename Type, auto = stripped_type_name<Type>().find_first_of('.')>
-[[nodiscard]] static constexpr std::string_view type_name(int) noexcept {
+[[nodiscard]] constexpr std::string_view type_name(int) noexcept {
     constexpr auto value = stripped_type_name<Type>();
     return value;
 }
 
 template<typename Type>
-[[nodiscard]] static std::string_view type_name(char) noexcept {
+[[nodiscard]] inline std::string_view type_name(char) noexcept {
```
skypjack, line 47: No need to make this one `inline`.

src/entt/core/type_info.hpp
```diff
@@ -38,26 +38,26 @@ template<typename Type>
...
-[[nodiscard]] static constexpr id_type type_hash(int) noexcept {
+[[nodiscard]] constexpr id_type type_hash(int) noexcept {
     constexpr auto stripped = stripped_type_name<Type>();
     constexpr auto value = hashed_string::value(stripped.data(), stripped.size());
     return value;
 }
 
 template<typename Type>
-[[nodiscard]] static id_type type_hash(char) noexcept {
+[[nodiscard]] inline id_type type_hash(char) noexcept {
```
skypjack, line 60: No need to make this one `inline`.

skypjack (changes requested):

DonutVikingChap:
Ok, the inline specifiers are removed now. In my own projects I usually mark all header-defined free functions either inline or constexpr even when they're templates. It's mostly for consistency, but also to make sure that the inline specifier is not accidentally forgotten in case the function is later refactored to not be a template, though I agree that it's not strictly necessary here.

skypjack:
Yeah, don't get me wrong. What you say makes perfectly sense. I'm just trying to stick with the coding style of this project. 👍 
Thanks for changing it. Let me know if you spot any other changes that are useful to make it work with C++20.
I also want to create a C++20 branch of the library soon, most likely after the vacations. 🥳

skypjack (approved):

PR #1069, opened 2023-09-18
Add organizer dependencies

cjhowedev:
Adds the indices of the in edges of organizer vertices, so that dependencies can be checked for completion before running a vertex's task. Also simplifies how the `is_top_level` field is calculated to reuse the passed in dependencies.


cjhowedev:
Yes, I can rename it if that helps clarify.

skypjack:
Well, I don't like `parent` either but I'd change the name to _<let's find the right one>_.
The fact is that _dependencies_ makes me think to something on which the node depends rather than something that depends on the node, if you get what I mean.

Innokentiy-Alaytsev:
1. Dependent
2. Successor
3. Derivative
4. Subject (?)

cjhowedev:
Hmm, if we don't use the terminology `parent`, I don't think we should use `children` either. I think the best path forward is to rename both to `successors`/`predecessors` to reduce confusion as to where they are in the graph.

skypjack:
Approved. 👍

cjhowedev:
Updated the commit message to match others. I also decided to keep `children` around for backwards compatibility, and have created `successors`/`predecessors`. The code should be ready for review.

src/entt/entity/organizer.hpp
```diff
@@ -178,11 +178,13 @@ class basic_organizer final {
          * @param vtype True if the vertex is a top-level one, false otherwise.
          * @param data The data associated with the vertex.
          * @param edges The indices of the children in the adjacency list.
+         * @param edges The indices of the dependencies in the adjacency list.
```
skypjack, line 181: `edges` -> `predecessors`?

src/entt/entity/organizer.hpp
```diff
@@ -178,11 +178,13 @@ class basic_organizer final {
          * @param vtype True if the vertex is a top-level one, false otherwise.
```
skypjack, line 178: `vtype` should go away, isn't it?

src/entt/entity/organizer.hpp
```diff
@@ -263,11 +265,27 @@ class basic_organizer final {
...
+         */
+        const std::vector<std::size_t> &successors() const noexcept {
+            return successors_vec;
+        }
+
+        /**
+         * @brief Returns the list of successors for a given vertex.
+         * @return The list of successors' indices for the vertex.
          */
         const std::vector<std::size_t> &children() const noexcept {
```
skypjack, line 287: Keep this one here is fine but I'd add a `[[deprecated("use ::successors")]]` too.

test/entt/entity/organizer.cpp
```diff
@@ -63,15 +63,30 @@ TEST(Organizer, EmplaceFreeFunction) {
     ASSERT_FALSE(graph[2u].top_level());
     ASSERT_FALSE(graph[3u].top_level());
 
-    ASSERT_EQ(graph[0u].children().size(), 2u);
```
skypjack, line 66: Since `children` still exists, please, do not drop its tests or the coverage tool will complain. 👍

skypjack (changes requested):

test/entt/entity/organizer.cpp
skypjack: Ping @cjhowedev 🙂

skypjack:
Available on `wip`, updated to the latest version of the library.

Issue #535, opened 2020-08-03
Sorting does not appear to be working


dgoad95:
```
struct ID
{
	ID( )
	{
		id += 1;
	}

	static int id;
};

int ID::id = 0;

entt::entity e[10];
for (int i = 0; i < _countof(e); i++)
{
	entity[i] = registry.create( );
	registry.emplace<ID>(entity[i]);
}

registry.destroy(entity[5]);
registry.sort<ID>([ ] (const auto& lhs, const auto& rhs)
{
	return lhs.id < rhs.id;
});
```

```
1>------ Build started: Project: Haven, Configuration: Debug Win32 ------
1>Editor.cpp
1>C:\Users\dg\Documents\Visual Studio 2013\Projects\Haven\entt\entity\sparse_set.hpp(518,41): warning C4244: 'argument': conversion from 'const __int64' to 'const int', possible loss of data
1>C:\Users\dg\Documents\Visual Studio 2013\Projects\Haven\entt\entity\registry.hpp(1276): message : see reference to function template instantiation 'void entt::sparse_set<Entity>::sort<Editor::Initialize::<lambda_2d85ba54190e27ad9ff36031c5c53ebe>,entt::std_sort,>(entt::sparse_set<Entity>::sparse_set_iterator,entt::sparse_set<Entity>::sparse_set_iterator,Compare,Sort)' being compiled
1>        with
1>        [
1>            Entity=entt::entity,
1>            Compare=Editor::Initialize::<lambda_2d85ba54190e27ad9ff36031c5c53ebe>,
1>            Sort=entt::std_sort
1>        ]
1>C:\Users\dg\Documents\Visual Studio 2013\Projects\Haven\entt\entity\registry.hpp(1279): message : see reference to function template instantiation 'void entt::sparse_set<Entity>::sort<Editor::Initialize::<lambda_2d85ba54190e27ad9ff36031c5c53ebe>,entt::std_sort,>(entt::sparse_set<Entity>::sparse_set_iterator,entt::sparse_set<Entity>::sparse_set_iterator,Compare,Sort)' being compiled
1>        with
1>        [
1>            Entity=entt::entity,
1>            Compare=Editor::Initialize::<lambda_2d85ba54190e27ad9ff36031c5c53ebe>,
1>            Sort=entt::std_sort
1>        ]
[...]

PR #1080, opened 2023-10-17
Docs - Fix typo in Sorting section

LiterallyVoid:
Using `const auto &...`  for the arguments doesn't work because the `entt::entity` overload has priority over the component (in this case, `renderable`) overload.

https://github.com/skypjack/entt/issues/535#issuecomment-667853342


skypjack:
Good catch. Thanks! 👍

PR #1098, opened 2023-12-13
build system: use env var for bazel version in CI

zaucy:
* `.bazelversion` wasn't being respected in CI
* also using 6.x instead of a specific version

when I get a chance I'll upgrade to 7 in a different PR


skypjack:
Thank you very much!!! ❤️

skypjack (approved):

PR #1107, opened 2024-01-20
build: don't install include/BUILD.bazel

### LINKED ISSUE #1104: 3.13.0 installs include/BUILD.bazel
Could you please remove it?

zaucy:
closes #1104


skypjack (approved):
I would have done it the other way around probably (as in `PATTERN "*.bazel" EXCLUDE`) but it's better this way actually. It holds in case we were to add a `foo.bar` file or similar in the future. 👍

PR #1124, opened 2024-03-13
build: update bazel common test target

zaucy:
~~other errors seem different. I'll have to take a look a little later. the CI is passing for   bazel windows though! 😅~~

~~edit: addeed 'common' to be considerd system inclues since they're using the `<` `>`~~

edit: since the 'common' files are at the root of the test module I cannot mark them as system headers with bazel so I opted to switch them to using quote includes. I'm hoping this is acceptable.


skypjack:
Oh, I see. I'll change something here but it's ok. Thanks! 👍

PR #1148, opened 2024-06-04
Bind might invalidate cpool reference if it mutates pools

iK4tsu:


iK4tsu:
This might deviate a bit from "you only pay for what you need" so perhaps there's an alternative solution

skypjack:
Fixed on the `wip` branch with a _proper_ change and a dedicated test. 👍

PR #1162, opened 2024-08-02
Fix typos in README

ChrisThrasher:
I believe the intended word here is "treatise" which is "a systematic exposition or argument in writing". A treaty is something different.


skypjack (approved):

skypjack:
Thanks!!

PR #1166, opened 2024-08-12
Enhance component_traits with advanced type traits and concepts

xNotTozic:


skypjack:
Hi, sorry for the late reply but I was out until yesterday.
Currently, EnTT requires C++17 and doesn't force C++20 on the final user.
I'm about to switch to this newer revision of the standard actually but I can't merge this PR upstream yet. I'm sorry.
On the other side, stay tuned and feel free to resubmit this PR as well as many others when I change the minimum standard for the library.
It would be great to have some help with the migration. 🙂

PR #1168, opened 2024-08-19
Automatic signal registration for components

### LINKED ISSUE #1165: Proposal: Automatic signal registration for components
Signals are a pretty useful tool offered by EnTT, especially when used with components. However, the fact that you have to explicitly specify each connection listener => signal might be not-optimal, especially for cases such as two registries, in which you need to connect listeners for both registries. Another problematic case is on game engines/libraries that expose components in their API. In this case it is not pretty clear when the user should attach the listeners to the signals and one might even forget to connect the listeners at all.

For this reason, I propose the sigh mixin storage should provide a mechanism to detect the signals of a component for which we are interested and connect the listeners for us. In code this would look like the following:
```cpp
struct dummy final
{
    static void on_construct(entt::registry &reg, entt::entity id)
    {
        fmt::print("Added `dummy` to {}\n", entt::to_entity(id));
    }

    static void on_update(entt::registry &reg, entt::entity id)
    {
        fmt::print("Updated `dummy` of {}\n", entt::to_entity(id));
    }

    static void on_destroy(entt::registry &reg, entt::entity id)
    {
        fmt::print("Removed `dummy` from {}\n", entt::to_entity(id));
    }

    int foo; // just to demonstrate `on_update`
};
```

In this specific case, the pool/storage would call `on_construct().connect<&dummy::on_construct>()` (the same for the other listeners) in the constructor of the storage itself, thus making it easier for the user to declare listeners for components and make sure we don't forget to attach them again :).

The change is simply isolated on the `basic_sigh_mixin` constructor, here's a sample [implementation](https://godbolt.org/z/GY5dqKWbj) (changes start on line 106). C++20 concepts are used for simplicity, but a C++17 impl. is possible.

For now these questions are still open:
- How should this interact with named pools? Do we want listeners attached to named pools as well or do we want them disabled?
- How do we opt in/out in named pools' case?

If you think this should be part of EnTT I can provide a PR with a C++17 implementation.

TerensTare:
This is the promised implementation for #1165, adding support for binding static functions of components to signals related to their lifetime. 

The functions should be named `on_construct`, `on_update` and `on_destroy`, be `static` to the component and have their signature as `void (entt::basic_registry<Entity, Alloc> &reg, Entity id)` to be considered by the default storage. The functions will be registered only for the specified registry specializations and only if the pool/storage of your component is the default storage. 

A component can implement any of these functions and doesn't need to implement all of them for the feature to work. Also note that these functions will also be available to the named pools of the component as well. If you don't want this behavior, you should have two component types: one exposing the signals and one that doesn't.

Here is an example component incorporating all the supported signals:

```cpp
#include <entt/entity/registry.hpp>
#include <fmt/core.h>

struct dummy final
{
    // the registry here can be a custom one as well (ie. with a custom entity/allocator)
    static void on_construct(entt::registry &reg, entt::entity id)
    {
        fmt::print("Added `dummy` to {}\n", entt::to_entity(id));
    }

    static void on_update(entt::registry &reg, entt::entity id)
    {
        fmt::print("Updated `dummy` of {}\n", entt::to_entity(id));
    }

    static void on_destroy(entt::registry &reg, entt::entity id)
    {
        fmt::print("Removed `dummy` from {}\n", entt::to_entity(id));
    }

    int foo; // just to demonstrate `on_update`
};
```

### DIFF
--- a/src/entt/entity/mixin.hpp
+++ b/src/entt/entity/mixin.hpp
@@ -11,6 +11,42 @@
 
 namespace entt {
 
+/*! @cond TURN_OFF_DOXYGEN */
+namespace internal {
+
+template <typename T, typename Reg, typename = void>
+struct has_on_construct final : std::false_type {};
+
+template <typename T, typename Reg>
+struct has_on_construct<
+    T, Reg,
+    std::void_t<decltype(T::on_construct(std::declval<Reg &>(),
+                                         std::declval<Reg &>().create()))>>
+    : std::true_type {};
+
+template <typename T, typename Reg, typename = void>
+struct has_on_update final : std::false_type {};
+
+template <typename T, typename Reg>
+struct has_on_update<
+    T, Reg,
+    std::void_t<decltype(T::on_update(std::declval<Reg &>(),
+                                      std::declval<Reg &>().create()))>>
+    : std::true_type {};
+
+template <typename T, typename Reg, typename = void>
+struct has_on_destroy final : std::false_type {};
+
+template <typename T, typename Reg>
+struct has_on_destroy<
+    T, Reg,
+    std::void_t<decltype(T::on_destroy(std::declval<Reg &>(),
+                                       std::declval<Reg &>().create()))>>
+    : std::true_type {};
+
+} // namespace internal
+/*! @endcond */
+
 /**
  * @brief Mixin type used to add signal support to storage types.
  *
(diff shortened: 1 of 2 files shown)


skypjack (commented):
LGTM. Just a couple of things actually. I'm not sure I do want a separate test file but I can move around things on my own later on, so let it be for now. However, it would be great if you could use the clang format embedded with the project before merging. The formatting seems different from that of the project but it could also be that you've already used it and I've missed some configuration. Is it the case?

TerensTare:
You're right actually, I did the changes and forgot to run clang-format before committing 😅, but it should be fixed now. Let me know if there's anything else.

skypjack:
Cool, thanks. Changed the target branch to `wip` and started all runs.
I've the last work-from-who-knows-where week of the year, then I'll be able to focus on all pending PRs and issues.
Thanks for your patience. August is a crazy month here, sorry for the lag. 😅

PR #1196, opened 2024-12-01
Joining views with storages

TerensTare:
Views are pretty much a core utility of EnTT already and there is good support for many common operations. One of them is joining with other views, especially useful when one view might have named pools. However, the syntax becomes a little noisy when you need to join a view with one storage, which most often is a named `void` storage or a storage out of the registry (from my experience). To do this, we currently have to do something like:
```cpp
entt::storage_for_t<void> dirty_entities;
// ...
auto view = registry.view<C1, const C2>() | entt::basic_view{dirty_entities};
for (auto id : view) {
    // ...
}
```

To address this, I propose to add a `join` function to `basic_view`, which works similarly to the existing `operator|`, except that it only accepts a storage as a parameter instead of a view. So the example before would now look like this:
```cpp
entt::storage_for_t<void> dirty_entities;
// ...
auto view = registry.view<C1, const C2>().join(dirty_entities);
for (auto id : view)
{
    // ...
}
```

Now the function returns a new view, so you can chain `.join()` calls one after another. This is something of preference but I would advise to use `operator|` in that case.

### DIFF
--- a/src/entt/entity/view.hpp
+++ b/src/entt/entity/view.hpp
@@ -647,10 +647,12 @@ public:
      * @note This function only supports adding one storage, use @ref operator| to join two views instead.
      * @sa operator|
      */
-    template<typename OGet>
-    [[nodiscard]] auto join(OGet &other) const noexcept {
-        return internal::view_pack<basic_view<get_t<Get..., OGet>, exclude_t<Exclude...>>>(
-            *this, basic_view<get_t<OGet>, exclude_t<>>{other}, std::index_sequence_for<Get...>{}, std::index_sequence_for<Exclude...>{}, std::index_sequence_for<OGet>{}, std::index_sequence_for<>{});
+    template<typename OGet, typename = std::enable_if_t<std::is_lvalue_reference_v<OGet>>>
+    [[nodiscard]] auto join(OGet &&other) const noexcept {
+        using oget_noref = std::remove_reference_t<OGet>;
+
+        return internal::view_pack<basic_view<get_t<Get..., oget_noref>, exclude_t<Exclude...>>>(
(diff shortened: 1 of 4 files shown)


skypjack:
Out of curiosity, why not:
```cpp
auto view = registry.view<C1, const C2>() | dirty_entities;
```
It seems more on the line of the already existing functionality. Any particular reasons for using a dedicate function instead?

TerensTare:
> Any particular reasons for using a dedicate function instead?

Purely a matter of semantics, since storages are a different type from views. Nothing against `operator|` except that. Also, I just realized that the implementation does not support joining a view with a const-storages,I will have to add that too.

src/entt/entity/view.hpp
```diff
@@ -638,6 +638,23 @@ class basic_view<get_t<Get...>, exclude_t<Exclude...>, std::enable_if_t<(sizeof.
...
+     * @brief Attaches a storage to the end of the view for iterating.
+     *
+     * @tparam OGet The type of the storage to attach to the view.
+     * @param other The storage to attach to the view.
+     * @return A new view with the given storage attached.
+     * @note This function only supports adding one storage, use @ref operator| to join two views instead.
+     * @sa operator|
+     */
+    template<typename OGet, typename = std::enable_if_t<std::is_lvalue_reference_v<OGet>>>
+    [[nodiscard]] auto join(OGet &&other) const noexcept {
```
skypjack, line 651: Ehy, sorry for the late reply. I've been busy recently.
I think you can have `OGet &other` and remove the `is_lvalue_reference_v` check as a consequence, no? `remove_reference_t` below is also useless then.
Same for the other `join`. Am I missing something? 👍

TerensTare: Huh I'd swear I had a problem with the `const` not propagating from storage to view earlier, I probably did something wrong... Anyway I just reverted the changes, let me know if you need anything else.

skypjack:
Looks like my playing with branches other than `master` messed up here. I'll merge the PR to a yet another branch and clean it up. 👍

PR #1201, opened 2024-12-12
Avoid pointless lookups in try_cast

morbo84:
We don't need a complete `meta_type` for the target type of the `try_cast`, but we are OK with the `type_info`.

### DIFF
--- a/src/entt/meta/node.hpp
+++ b/src/entt/meta/node.hpp
@@ -198,8 +198,8 @@ template<typename... Args>
     return value(context);
 }
 
-[[nodiscard]] inline const void *try_cast(const meta_context &context, const meta_type_node &from, const meta_type_node &to, const void *instance) noexcept {
-    if((from.info != nullptr) && (to.info != nullptr) && *from.info == *to.info) {
+[[nodiscard]] inline const void *try_cast(const meta_context &context, const meta_type_node &from, const type_info *to, const void *instance) noexcept {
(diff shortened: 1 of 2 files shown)


src/entt/meta/node.hpp
```diff
@@ -198,8 +198,8 @@ template<typename... Args>
     return value(context);
 }
 
-[[nodiscard]] inline const void *try_cast(const meta_context &context, const meta_type_node &from, const meta_type_node &to, const void *instance) noexcept {
-    if((from.info != nullptr) && (to.info != nullptr) && *from.info == *to.info) {
+[[nodiscard]] inline const void *try_cast(const meta_context &context, const meta_type_node &from, const type_info *to, const void *instance) noexcept {
+    if((from.info != nullptr) && (to != nullptr) && *from.info == *to) {
```
skypjack, line 202: I wonder if we can make it a `const type_info &` instead.
`can_cast` is the only one that could pass a `nullptr` here, but we can prevent it evantually.
`try_cast` would benefit from less checking instead.

skypjack (commented):
Quick comment/question before merging.
If it's tricky to do, ignore it and I'll look into that part later after the merge. 👍

src/entt/meta/node.hpp
morbo84: Good idea 👍

skypjack (approved):
Ignore the failure. GH is switching the `ubuntu-latest` identification and this causes troubles.

Issue #1213, opened 2025-01-25
g++17 meta.hpp compilation error (fix proposal)


acilione:
Compiling with g++ -std=c++17 on WLS (Ubuntu) I was getting the following error:

```
In file included from ./entt/meta/container.hpp:19,
                 from ./entt/entt.hpp:44,
                 from main.cpp:4:
./entt/meta/meta.hpp: In member function ‘entt::meta_func entt::meta_func::next() const’:
./entt/meta/meta.hpp:1041:81: error: use of deleted function ‘constexpr entt::meta_func::meta_func()’
 1041 |         return (node.next != nullptr) ? meta_func{*ctx, *node.next} : meta_func{};
      |                                                                                 ^
./entt/meta/meta.hpp:953:5: note: ‘constexpr entt::meta_func::meta_func() noexcept’ is implicitly deleted because its exception-specification does not match the implicit exception-specification ‘noexcept (false)’
  953 |     meta_func() noexcept = default;
      |     ^~~~~~~~~
./entt/meta/meta.hpp: In member function ‘entt::meta_type entt::meta_type::template_type() const’:
./entt/meta/meta.hpp:1288:134: error: use of deleted function ‘constexpr entt::meta_type::meta_type()’
 1288 |         return (node.templ.resolve != nullptr) ? meta_type{*ctx, node.templ.resolve(internal::meta_context::from(*ctx))} : meta_type{ 
;
      |
./entt/meta/meta.hpp:1132:5: note: ‘constexpr entt::meta_type::meta_type() noexcept’ is implicitly deleted because its exception-specification does not match the implicit exception-specification ‘noexcept (false)’
 1132 |     meta_type() noexcept = default;
      |     ^~~~~~~~~
./entt/meta/meta.hpp: In member function ‘entt::meta_type entt::meta_type::template_arg(entt::meta_type::size_type) const’:
./entt/meta/meta.hpp:1297:130: error: use of deleted function ‘constexpr entt::meta_type::meta_type()’
 1297 |         return index < template_arity() ? meta_type{*ctx, node.templ.arg(internal::meta_context::from(*ctx), index)} : meta_type{};   
      |                                                                                                                                  ^    
./entt/meta/meta.hpp: In member function ‘entt::meta_data entt::meta_type::data(entt::id_type) const’:
./entt/meta/meta.hpp:1344:71: error: use of deleted function ‘constexpr entt::meta_data::meta_data()’
[...]

PR #1214, opened 2025-01-25
Fix issue https://github.com/skypjack/entt/issues/1213#issue-2810640371

acilione:
All tests passed

image

### DIFF
--- a/src/entt/meta/meta.hpp
+++ b/src/entt/meta/meta.hpp
@@ -827,7 +827,7 @@ struct meta_data {
     using size_type = typename internal::meta_data_node::size_type;
 
     /*! @brief Default constructor. */
-    meta_data() noexcept = default;
+    meta_data() noexcept(false) = default;
 
     /**
      * @brief Context aware constructor for meta objects.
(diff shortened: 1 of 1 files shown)


skypjack:
I see your concern. However, I don't think that making all constructors `noexcept(false)` because of a faulty compile is the way to go. We can make them conditionally `noexcept` and have the _good_ compilers take advantage of it.

acilione:
Sorry, only now I had time to look at it.

Do you think something like would be clean enough? 

#if defined(__GNUC__) && !defined(__clang__)
    #define NOEXCEPT noexcept(false)
#else
    #define NOEXCEPT noexcept
#endif

I made a little trial program to ensure that NOEXCEPT = noexcept(false) only when g++ is used

skypjack:
I think the last proposal suffers from the same problem actually. What version of the compiler causes the problem exactly? I can try to pack a workflow on the CI to make some tests.

acilione:
Hi, here is the output of g++ --version: 

g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0

skypjack:
9? 😲 That's... old, at least. 
Honestly, I'm not so sure about changing the codebase to support such an old compiler.
Does the problem occur with more modern versions?

skypjack:
Sorry. The proposed solution makes things worse and I don't think it's worth it for such an old compiler. I'm closing the PR.

Issue #1222, opened 2025-02-18
signals (sigh_mixin) with non-public inherited registry


actondev:
Reopening the discussion of https://github.com/skypjack/entt/issues/1078 (and https://github.com/skypjack/entt/pull/1079 )
While e3475c026 makes it possible to use custom inherited registries with the sigh_mixin, it only works with public inheritance.

With the attached diff (which uses protected inheritance in the custom registry test), the sigh functionality **breaks** (see below, tried on top of v3.14.0)

Our usecase for using protected inheritance, is to disable any mutation that doesn't go through the `patch` public API, ie that doesn't trigger signals. (f.ex. via a non-const registry view `registry.view<Component,..>(..)` the components can be mutated)

The problem seems to be resolved by storing the actual inherited `registry_type` in the `basic_sigh_mixin` (instead of the `basic_registry_type`). Can gladly prepare a PR if you agree with the approach, but wanted to discuss the issue first.

---

The diff that breaks the custom registry test. The issue comes from casting
```
entt/src/entt/entity/mixin.hpp:46:19: error: cannot cast 'test::basic_custom_registry<test::entity>' to its protected base class 'entt::basic_registry<test::entity>'
   46 |             reg = any_cast<Type>(&value);
```

``` diff
diff --git a/test/common/registry.h b/test/common/registry.h
index 55db7f1fd..400cd447e 100644
--- a/test/common/registry.h
+++ b/test/common/registry.h
@@ -6,7 +6,27 @@
 namespace test {
 
 template<typename Entity>
-struct basic_custom_registry: entt::basic_registry<Entity> {};
+class basic_custom_registry: protected entt::basic_registry<Entity> {
+    using registry_type = entt::basic_registry<Entity>;
+
+    // Guarded custom registry that exposes a subset of basic registry methods
+    // !!! DOES NOT WORK !!!
+  public:
+    using entity_type = registry_type::entity_type;
+    using traits_type = entt::entt_traits<Entity>;
+    using base_type = entt::basic_sparse_set<Entity>;
+    using allocator_type = registry_type::allocator_type;
+
+    // any exposed methods need to be forwarded
+
+    [[nodiscard]] Entity create() { return registry_type::create(); }
+
+    template<typename Type, typename It, typename... Args>
+    void insert(It begin, It end, Args &&... args)
+    {
+        registry_type::template insert<Type>(begin, end, std::forward<Args>(args)...);
+    }
+};
 
 } // namespace test
 
```

[...]

PR #1225, opened 2025-02-20
mixins working with custom protected registry

actondev:
Working but.. not-confident-about-the-whole-thing addressing https://github.com/skypjack/entt/issues/1222
The issue is I don't quite like this reinterpret_cast there

What I don't quite understand, is that both the if statements get hit
- when constructing the custom registry, the first if statement gets hit, this is on the `basic_registry` constructor when calling `rebind`. How come any_cast detects the custom registry?
- when creating storages from `assure` (let's say from `.emplace<Component>` from the custom registry, assuming it was exposed), the newly created pool binding hits the second if and the reinterpret_cast is needed.

### DIFF
--- a/src/entt/entity/mixin.hpp
+++ b/src/entt/entity/mixin.hpp
@@ -130,7 +130,12 @@ private:
     }
 
     void bind_any(any value) noexcept final {
-        owner = internal::any_to_owner<registry_type>(value);
+        if(auto *registry = any_cast<owner_type>(&value)) {
+            owner = registry;
+        }
+        else if(auto *base_registry = any_cast<basic_registry_type>(&value)) {
+            owner = reinterpret_cast<owner_type *>(base_registry);
+        }
         underlying_type::bind_any(std::move(value));
     }
 
@@ -386,7 +391,7 @@ public:
     }
 
 private:
-    basic_registry_type *owner;
+    owner_type *owner;
     sigh_type construction;
     sigh_type destruction;
     sigh_type update;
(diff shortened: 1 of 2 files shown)


test/common/registry.h
```diff
@@ -5,8 +5,37 @@
 
 namespace test {
 
+/// Custom registry that exposes a subset of basic registry methods
 template<typename Entity>
-struct basic_custom_registry: entt::basic_registry<Entity> {};
+class basic_custom_registry: protected entt::basic_registry<Entity> {
+    using registry_type = entt::basic_registry<Entity>;
```
skypjack, line 11: There is a clang-format file in EnTT. Please, use it also for these changes. 🙏🙂

src/entt/entity/mixin.hpp
```diff
@@ -130,7 +130,12 @@ class basic_sigh_mixin final: public Type {
     }
 
     void bind_any(any value) noexcept final {
-        owner = internal::any_to_owner<registry_type>(value);
+        if(auto *registry = any_cast<owner_type>(&value)) {
```
skypjack, line 133: I'd rather change the code of `any_to_owner` than duplicate the logic in two places.
Also, I agree that `reinterpret_cast` is a no-go. Probably a `static_cast` is more than enough and slightly safer.

skypjack (changes requested):

skypjack:
I dediced not to merge this PR, because I don't like the `reinterpret_cast` nature of the solution.
Let's move the discussion back to the issue. I think a friend declaration is more than enough actually.

PR #1244, opened 2025-04-03
SFINAE friendly comparison operators

ZXShady:
Hi, I had an issue with an assert aware expression decomposer where the expression decomposer uses overloaded `operator==` and `operator!=` without any constrains on the second type of parameter which conflicted with `entt;:null_t` when I wanted to override `ENTT_ASSERT` with `SUPER_COOL_ASSERT` macro


src/entt/entity/entity.hpp
```diff
@@ -219,7 +226,7 @@ struct null_t {
      * @param other A null object.
      * @return True in all cases.
      */
-    [[nodiscard]] constexpr bool operator==([[maybe_unused]] const null_t other) const noexcept {
```
skypjack, line 222: These names are there for documentation purposes.
Please, revert all changes that are not strictly necessary to show your intent.
Otherwise, one must filter the changes that are irrelevant to the review to get the full picture.
Thanks in advance.

ZXShady: You are right

ZXShady:
> I introduced the `entity_like` concept and tried to constrain `null_t` and `tombstone_t`. In theory, your issue should be solved in a more C++20-ish way now. Can you double check branch `wip` to confirm it? Or provide a repro for that eventually. I had a similar problem with the `handle` class, and the new concept along with the constrained traits solved it transparently.


Dang, github didn't notify at all of your reply, I will test this in a later time . sorry for delay

ZXShady:
@skypjack , sadly this doesn't fix it infact the 'entity_like' concept does nothing actually

```cpp
namespace internal {
template<typename>
struct entt_traits;
} // namespace internal
template<typename Type>
concept entity_like = requires {
    typename internal::entt_traits<Type>; // this is always ture as internal::entt_traits is always here as a symbol, you want to check something sfinaeable like ::value_type of it 
};
static_assert(entity_like<void>) // compiles and returns true

// "correct"
template<typename Type>
concept entity_like = requires {
    typename internal::entt_traits<Type>::value_type; 
};
```

I said correct in qoutes, because it causes another issue now in the fwd.hpp file it fails to compile the aliases

```cpp
template<entity_like Entity = entity, typename = std::allocator<Entity>>
class basic_registry;

// entity here has no entt_traits definition avaiable at this point and fails to compile, we don't want that we want at point of use
using registry = basic_registry<>; 
```

Sure you could inside fwd.hpp header declare the `entt_traits` for `entity` but this is an unnecessary compile time cost the better fix is not using concepts at all and usea  normal static_assert like you did in the olDC++17, the concept bring nothing of value, it has worse error messages than a static_assert and causes slower compilation.



the fix

```cpp
// Before
template<entity_like Entity, typename Allocator>
class basic_registry {
...
};

// After
template<typename Entity, typename Allocator>
class basic_registry {
    static_assert(entity_like<Entity>,"Must satisfy entity traits"); // you can even put a message, concepts can't
    ...
};
```

in my opinion, the use of concepts should reverted UNLESS you need subsituation failure. you should use `static_assert`s, they bring

1. custom error messages
2. shorter symbols (helpful when parsing the insane errors I get from entt)
3. faster compilation

but in places like operators and anything supposed to be found by adl concepts should be used to constraint the arguements. but member functions and classes should just continue using static_asserts

skypjack:
> infact the 'entity_like' concept does nothing actually

@ZXShady 🤦‍♂️ legacy of the first attempt, I didn't realize I left it like this.
Should be updated on the `wip` branch now, and also scoped to avoid the _second_ issue you had with `fwd.hpp`.
Let me know. I tried your test and it passed without problems. Thanks.

ZXShady:
I think you should add a static_assert inside the class, because you removed the concept from the typename parameter, I went ahead and installed wip branch and tried it and it works nicely.


Thanks jack.

skypjack:
> I think you should add a static_assert inside the class, because you removed the concept from the typename parameter

Which class exactly? I removed the concept from a lot of them... 😅

PR #1254, opened 2025-05-14
Fix Ninja not beeing found on Windows testbed Ci

Fancy2209:
Should fix the issue that made the following run fail:
https://github.com/skypjack/entt/actions/runs/14351371989

### DIFF
--- a/.github/workflows/testbed.yml
+++ b/.github/workflows/testbed.yml
@@ -59,6 +59,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+
+    - name: Install Ninja
+      run: winget install --id=Ninja-build.Ninja  -e
+    
     - name: Compile testbed
       working-directory: build
       run: |


Fancy2209:
Oh wait I just realized, windows-latest is still 2022 and winget is only available on windows-2025, I'll fix this to another method when I can, my bad

Fancy2209:
Oh no it seems the readme for the actions containers is just outdated and it is windows-2025?

skypjack:
The `testbed` workflow is is fixed already on the `wip` branch. See [here](https://github.com/skypjack/entt/actions/workflows/testbed.yml). Is this what you're trying to do? In this case, we can safely close the PR as _done_ 👍

Fancy2209:
> The `testbed` workflow is is fixed already on the `wip` branch. See [here](https://github.com/skypjack/entt/actions/workflows/testbed.yml). Is this what you're trying to do? In this case, we can safely close the PR as _done_ 👍

Oh yep, my bad

Issue #658, opened 2021-02-14
Support using entt::type_info as key in associative containers


jonesmz:
I want to use `entt::type_info` as the key in an associative container (e.g. `std::map`). However, `type_info` has some problems.

```cpp
template<typename Type>
type_info type_id() ENTT_NOEXCEPT;
```

could be constexpr if they didn't include type_seq. My knowledge with constexpr seems to be lacking here, so it's not clear that it's possible to do this at all, so maybe this aspect of the issue can be ignored.

That being said, the type_id() function should be [[nodiscard]]. As a (mostly) pure function, there's no reason to call it without doing something with the result. The default implementation being constexpr (if possible) and [[nodiscard]] wouldn't influence explicit specializations that the end user makes, so probably best not to rely on that inside of EnTT, but if the user of EnTT doesn't implement non-constexpr specializations of type_id<>() then they could take advantage of that. (Again, assuming it's even possible to use constexpr here, which it may not be)

------------------------

`type_seq`, `type_hash` and `type_name` should have implicit conversion operators to allow for short-hand usage.

```cpp
template<typename Type, typename = void>
struct type_seq final {
    [[nodiscard]] static constexpr id_type value() ENTT_NOEXCEPT;
    constexpr operator id_type() const ENTT_NOEXCEPT { return value(); }
};

template<typename Type, typename = void>
struct type_hash final {
    [[nodiscard]] static constexpr id_type value() ENTT_NOEXCEPT;
    constexpr operator id_type() const ENTT_NOEXCEPT { return value(); }
};

template<typename Type, typename = void>
struct type_name final {
    [[nodiscard]] static constexpr std::string_view value() ENTT_NOEXCEPT;
    constexpr operator std::string_view() const ENTT_NOEXCEPT { return value(); }
};
```

Use like:

```cpp
void foo(id_type);
foo(entt::type_seq<MyType>{});
foo(entt::type_seq<MyType>::value()); // same as above.
foo(entt::type_hash<MyType>{});
foo(entt::type_hash<MyType>::value()); // same as above.

void bar(std::string_view);
foo(entt::type_name<MyType>{});
foo(entt::type_name<MyType>::value()); // same as above.
```

[...]

PR #1334, opened 2026-04-22
Add std::string_view constructor to hashed_string

skhaz:
## Summary

Adds a convenience constructor accepting `std::string_view` directly, eliminating the need for manual `.data()` and `.size()` calls.

## Motivation

As noted in PR #824, `std::string_view` support was mentioned as desirable but not included at the time. This PR addresses that gap.

Many codebases use `std::string_view` extensively, and users currently write verbose code:

```cpp
// Before: verbose
const auto key = entt::hashed_string{sv.data(), sv.size()};

// After: clean and intuitive
const auto key = entt::hashed_string{sv};
```

## Implementation

```cpp
constexpr basic_hashed_string(std::string_view sv) noexcept
    : basic_hashed_string{sv.data(), sv.size()} {}
```

**Zero overhead**: Single conversion to `string_view`, then direct delegation.

**SFINAE constraint**: Excludes C-string pointers to avoid ambiguity with `const_wrapper` used by UDL operators (`_hs`, `_hws`).

## Test Coverage

- Null-terminated `string_view`
- Sized `string_view` (prefixes)
- Substring in middle of buffer
- Empty `string_view`
- Single character
- `string_view` with embedded nulls
- All 184 existing tests pass

## Checklist

- [x] Compiles without warnings (C++17)
- [x] All 184 core tests pass
- [x] Zero overhead implementation
- [x] Backwards compatible
- [x] Comprehensive test coverage

## Related

- PR #824: Store size information in hashed string
- Issue #658: Uses `view.data(), view.size()` pattern

/cc @skypjack

### DIFF
--- a/src/entt/core/hashed_string.hpp
+++ b/src/entt/core/hashed_string.hpp
@@ -3,6 +3,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <string_view>
 #include "fwd.hpp"
 
 namespace entt {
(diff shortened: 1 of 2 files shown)


skhaz:
Any chance to get a review on this?

skypjack:
The `string_view` constructor existed and was removed [here](https://github.com/skypjack/entt/commit/f984ae4930b0a1c99b5a6a9762adf303e5b5c0e9) to eliminate an unnecessary dependency. Support for the `char *` and size pair is sufficient and works perfectly for `string_view`s, as well as many other types.

skypjack:
My two cents: forcing an unnecessary dependency on all users isn't ideal, but it's easy to write a userspace shortcut in this case. I'd go with the latter. 👍

PR #1337, opened 2026-06-21
Caio/modules

caiowakamatsu:
This PR is for finish the work @MilanBorovy started with #1290 to get C++ 20 modules working with EnTT.

I will rebase it against `wip` due to the STL injection changes causing a massive merge conflict. 

---
### Original PR text
- Make EnTT consumable as C++ module
- Added ENTT_MODULE CMake option enabling EnTT compilation as C++ 
- Added ENTT_USER_CONFIG CMake option to be able to specify C++ config defines

I've [@MilanBorovy] noticed that similar pull rq is already ongoing but seems to be dead for a few months.

Instead of exporting symbols in separate files, export macros are used (similar to how fmt does it).

By setting `ENTT_MODULE` CMake option to ON the `EnTT::EnTT` target becomes module library allowing `EnTT` to be consumed as `import entt;`.

Another option `ENTT_USER_CONFIG` is added to `CMake`. It can be used to configure entt as on can't simply #define all configurations inside source with modules.

Usage example:

entt_config.hpp
```
#define ENTT_ID_TYPE std::uint64_t
#define ENTT_USE_ATOMIC
```

CMakeLists.txt
```
...
set(ENTT_MODULE ON)
cmake_path(APPEND PROJECT_SOURCE_DIR user_config.hpp OUTPUT_VARIABLE ENTT_USER_CONFIG)

add_subdirectory(entt)

...

target_link_libraries(exe PUBLIC EnTT::EnTT)
```

Some explanation of this would be nice to put in [wiki](https://github.com/skypjack/entt/wiki/Configuration), unfortunately I [@MilanBorovy] can't pull request changes to wiki AFAIK.


caiowakamatsu:
In favor of keeping this PR well scoped, I think an attempt to reconcile STL injection + modules should not be made here. Instead a separate PR should be done for that functionality.

skypjack:
> In favor of keeping this PR well scoped, I think an attempt to reconcile STL injection + modules should not be made here. Instead a separate PR should be done for that functionality.

That's fair. We still need to fix the conflicts though, and thanks for pointing this to `wip`. 🙏 
My idea was to cut a release by the end of the month, but I'm fine with doing it in July if we want to include this one.

caiowakamatsu:
@skypjack 
> My idea was to cut a release by the end of the month, but I'm fine with doing it in July if we want to include this one.

I don't think this is going to be able to be release-able in the next ~week~ in all likelihood. I've fixed a lot of the merge conflicts, but now there's a lot of compiler errors. I think getting this done by July may be realistic. But it's up to you if you want to hold off on the release for it :). I am ok with this either way

caiowakamatsu:
@alexpanter @JonasGilg 

Pinging you two to let you know that this branch should be stable / usable now for modules, since you guys showed interested in #1290.

 I have tested it on my machine with GCC 16.1.1 / CMake 4.4.

JonasGilg:
> @alexpanter @JonasGilg 
> 
> Pinging you two to let you know that this branch should be stable / usable now for modules, since you guys showed interested in #1290.
> 
>  I have tested it on my machine with GCC 16.1.1 / CMake 4.4. 

Thank you very much. Unfortunately I reverted the use of modules in my project, since there were too many bugs in MSVC with VS2022. I will visit this again when I can upgrade to VS2026.

skypjack:
> I've fixed a lot of the merge conflicts, but now there's a lot of compiler errors. I think getting this done by July may be realistic. But it's up to you if you want to hold off on the release for it :). I am ok with this either way

Not a big deal. Let's see how it goes. I started a run on the CI to know _where_ we are. 👍

skypjack:
@caiowakamatsu I think I'm missing something. How do we proceed with this? Any news?

caiowakamatsu:
Work has been really busy, I should be able to pick this up in the next few days. Apologies for the delay 😅

caiowakamatsu:
I got CI running on my fork, I plan to finish this up today if all things go well. If not EOW. Sorry for the massive delay y'all. This is my first major open source contribution 😅

caiowakamatsu:
@skypjack All the workflows look green for me, mind doing a (hopefully) final run?

The only one I can't run is codecov since I don't have the secrets required.

skypjack:
Cool cool. It works fine as far as I can tell! 🥳 
Since this PR contains many changes, and some are good to go even without the module stuff, I'd like to _deliver_ a bunch of them separately.
Do you mind if I _steal_ the relevant changes from your PR directly, and rebase it before merging the module related part?
The goal is to avoid asking you to split them yourself, so you don't waste time. Makes sense?

skypjack:
As an example, the `static constexpr`s turned into `inline constexpr`s are fixes that have nothing to do with module support, and they are good to go on `main` on their own. Similarly, I see that you turned a lot of functions into `friend` functions. Even if it's something somewhat required for module support, these changes can stay on their own commit. Speaking of which, can I ask you why it's required too? Mostly curiosity, I'm not opposed to this change.

CrackedMatter:
> [...] `static constexpr`s turned into `inline constexpr`s are fixes that have nothing to do with module support [...]

Declarations with internal linkage cannot be exported (https://eel.is/c++draft/module.interface#3).

skypjack:
Is there any way to add a non-regression test to be sure that I'm not going to break this stuff with future changes?

caiowakamatsu:
I think there might be some way, I'll have to think about this for a bit. Do we want to merge this in now? Getting regression tests in a later PR?

